Issue #23, opened 2016-03-21
marbl_interface_types should use marbl_log_type


mnlevy1981:
Several routines in `marbl_interface_types` print error messages directly to stdout and have comments about needing to trigger an abort... these should all be using `marbl_log_error`, which means they need to be passed a `marbl_log_type`

mnlevy1981:
Fixed in marbl0.3.0

PR #38, opened 2016-03-22
Bugfix/use status log

mnlevy1981:
This branch resolves issue #23 -- marbl_interface_types.F90 now provides output to the MARBL status log rather than writing to stdout (allowing errors to be trapped and returned to the GCM as well).

Ran the aux_pop_MARBL test suite, and all 9 tests were bit-for-bit with MARBL0.0.0


klindsay28:
For src/marbl_diagnostics_mod.F90, github states

1,805 additions, 227 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

I don't know how to do this.

klindsay28:
The string "error code returned" confuses me. When I see that, I'm expecting to see a code, but there is none. Perhaps I'm not understanding the semanatics of marbl_logging.F90 or how it is to be understood. Unfortunately, examining the source code of marbl_logging.F90 is helping me out. It would be useful if there were documentation at the top of marbl_logging.F90 describing usage.

mnlevy1981:
@klindsay28 -- to see all the diffs locally, do the following:
1. Add my fork as a remote to your local clone of  MARBL
   
   `$ git remote add mlevy git@github.com:mnlevy1981/MARBL.git`
2. Make all the branches / tags from the mlevy remote available locally (this will only update your `.git/` directory, it will not change any of the files outside that directory; i.e. it will not alter any of the files in your `src/` directory)
   
   `$ git fetch mlevy`
3. Compare the revision I branched from (`bce81dc`) to my `mlevy/bugfix/use_status_log` branch
   
   `$ git diff bce81dc mlevy/bugfix/use_status_log`

---

Instead of "error code returned ..." I can change the text to "error reported from [module]". I wasn't sure of the best way to word the messages, but consider the situation where the GCM calls MARBL_A(), which calls MARBL_B(), which calls MARBL_C(), which reports an error.

Ideally, the message log would show the following (in this order):

```
MARBL_C: [description of error]
MARBL_B: [tell user error came from call to MARBL_C()]
MARBL_A: [tell user error came from call to MARBL_B()]
GCM: [tell user error came from call to MARBL_A()]
```

The GCM would write the log to stdout or a file and then call `MPI_Abort()`. Like I said, I wasn't sure of the best way to actually word the error messages and am happy to change the text.

I'm happy to add documentation to `marbl_logging.F90` describing this usage.

klindsay28:
I'm concerned about using a module variable in marbl_logging.F90 as a temporary for message creation. It smells like a potential thread race condition waiting to bite us. Our calls to marbl_init, we are calling it for each thread, are in an iblock loop in ecosys_driver_init. If someone were to put on openmp do directive around the loop, say to speed things up, I think we'd intermittently be in trouble.

Would it be better for users of marbl_logging to either use their own temp string, or to move the temp string into one of the marbl_logging data types?

mnlevy1981:
I didn't think about the threading / race issues, but that's a great point. The safest fix would be to make `error_msg` and `status_msg` local variables in every routine that uses the status log. I'll probably just use one variable (`log_msg` or `log_message`) and pass it to both `log_error()` and `log_noerror()`.

mnlevy1981:
This is what the log looks like when I make `num_forcing_diags` too small:

```
 MARBL ERROR (marbl_interface_types:marbl_diagnostics_add): not enough memory al
 located for this number of diagnostics!
 MARBL ERROR (marbl_diagnostics_mod:marbl_diagnostics_init): Error reported from
  diags%add_diagnostic(O2SAT)
 MARBL ERROR (marbl_interface:marbl_init): Error reported from marbl_diagnostics
 _init()
 MARBL ERROR (ecosys_driver:ecosys_driver_init): error code returned from marbl(
 1)%init()
```

Obviously I need to update the "error code returned" message in the driver, but otherwise I think this is a pretty informative set of messages.

mnlevy1981:
Add when I make `num_surface_forcing_fields` too small:

```
 MARBL ERROR (marbl_interface_types:marbl_forcing_fields_add): not enough memory
  allocated for number of forcing fields!
 MARBL ERROR (marbl_mod:marbl_sflux_forcing_fields_init): Error reported from fo
 rcing_fields%add(xco2_alt_co2)
 MARBL ERROR (marbl_interface:marbl_init): Error reported from marbl_init_surfac
 e_forcing_fields()
 MARBL ERROR (ecosys_driver:ecosys_driver_init): error code returned from marbl(
 1)%init()
```

Again, I need to clean up the error message coming from the driver.

@klindsay28 -- at this point, I think I've addressed everything you've brought up today. I also brought this branch up to date with master (which was updated when I brought in the `surface_forcing_output` branch), so to see the full set of diffs you need to run

```
$ git fetch mlevy
$ git diff mlevy/master mlevy/bugfix/use_status_log
```

If there's more work to be done on this branch, I'll put it off until after I finish the tracer index clean-up... but I think this is pretty close.

klindsay28:
subname in subroutines of marbl_co2calc_mod.F90 are not consistent with the full module name and/or the subroutine name, or how many colons are used to separate the module name from the subroutine name. Suggested changes are:
-    character(*), parameter :: subname = 'marbl_co2calc:marbl_co2calc_surf'
-    character(*), parameter :: subname = 'marbl_co2calc_mod:marbl_co2calc_surf'
-    character(*), parameter :: subname = 'marbl_co2calc:marbl_comp_CO3terms'
-    character(*), parameter :: subname = 'marbl_co2calc_mod:marbl_comp_CO3terms'
-    character(*), parameter :: subname = 'marbl_co2calc:marbl_comp_htotal'
-    character(*), parameter :: subname = 'marbl_co2calc_mod:comp_htotal'
-    character(*), parameter :: subname = 'marbl_co2calc::drtsafe'
-    character(*), parameter :: subname = 'marbl_co2calc_mod:drtsafe'

I now see that this is the case in numerous other instances across MARBL source code. Running the command
grep -Ei 'subname _=|^ *subroutine|^ *function' src/_.F90 | grep -B1 'subname *='
picks up problems in marbl_restore_mod.F90, marbl_mod.F90, marbl_diagnostics_mod.F90, and marbl_ciso_mod.F90

Perhaps fixing these mismatches in files not already changed on this branch could be postponed to another tag/checkin. Perhaps a bug should be filed for this.

src/marbl_logging.F90
```diff
@@ -1,5 +1,51 @@
 module marbl_logging
```
mnlevy1981, line 1: Notes from a conversation with @klindsay28 

1) Documentation at top of routine would be more readable in a bullet-point format
- Routines used to write log entries
- Routines used by the driver
  -- Include psuedo-code for marl_log_print!

2) Documentation in marbl_log_[noerror|error|error_trace] should define CodeLoc and whatnot
"CodeLoc is the name of the subroutine that has called marbl_log_error_trace, RoutineName is the name of the routine that was called by CodeLoc and now labort_marbl is true"

3) horizontal rule between subroutines (!====================)

4) Need better documentation on OutputOptions (mention that it is included in the code as a hook for future development), remove namelist reference from comment at top of marbl_log_error

klindsay28:
everything else looks fine to me
FYI, I learned from google that adding ?w=1 to the URL when examining files changed will show the diff with whitespace ignored (it mostly works in my experience)

mnlevy1981:
Note that this update also requires an update to `ecosys_driver.F90` (which currently uses two variables from `marbl_logging` that have been removed from the module); I realize that I haven't committed those changes to a branch of POP yet, so I need to do that and then I'll post a link with the POP changes + test results.

mnlevy1981:
The POP branch needed to run with this MARBL tag is

https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_status_log_tags/marbl_dev_status_log_n01_marbl_dev_n03_cesm_pop_2_1_20160407/

on yellowstone_intel, all aux_pop and aux_pop_MARBL passed and also passed baseline comparisons (using cime 4.4.8 and cice 20160309; baseline POP was the March 17 trunk tag)

PR #39, opened 2016-03-24
Enhancement/nag compatibility

mnlevy1981:
Introduced a new script to make sure no lines exceed NAG's 132 character limit, and updated the Makefile in bld-test/ to run this script as part of the build - so a developer without access to the NAG compiler still has an easy way to make sure his / her changes to the code will be parsed correctly by NAG.

I also updated the logic in the Makefile to make it a little easier to add new compilers -- there is a single `$(LIB_DIR)/%.a` target and a bunch of `PHONY` targets that set the compiler / flags before building the library.


mnlevy1981:
I don't think we want to accept this pull request -- I've talked with some CSEG folks, and a better solution seems to be setting up a new repository for tools like "check_line_length.sh" that we will want to use across a wide range of CESM-supported repositories. I'm closing this ticket while leaving my branch open so I can copy the Makefile (probably moving it to the write_namelist branch) and still have a copy of the scripts I want to eventually move to a new repository... which smells like a weekend project to me.

PR #44, opened 2016-04-02
Enhancement/surface forcing output

mnlevy1981:
I have generalized the surface forcing output data type, which required an updated POP branch as well (because the driver interface changed). See

https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n02_marbl_dev_n01_cesm_pop_2_1_20160317b

For the driver-side changes. I can think of one small tweak to make to this branch before accepting the pull request: we may want to print a log message when MARBL adds a new surface forcing outputs (right now the constructor logs errors, but that is it).

### DIFF
--- a/src/marbl_diagnostics_mod.F90
+++ b/src/marbl_diagnostics_mod.F90
@@ -50,6 +50,7 @@ module marbl_diagnostics_mod
   use marbl_interface_types , only : marbl_saved_state_type
   use marbl_interface_types , only : marbl_interior_forcing_input_type
   use marbl_interface_types , only : marbl_surface_forcing_output_type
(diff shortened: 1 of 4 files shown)


mnlevy1981:
I should add that this tag (as used in the updated POP tag) is bit-for-bit with the trunk for the `aux_pop_MARBL` test suite.

mnlevy1981:
Looks like I need to figure out how to provide a better commit log when accepting a pull request (I used the "squash" option to bring my 5 commits in as 1, and the commit message contains no information other than the name of the branch that got merged)

Issue #43, opened 2016-03-28
Sort out tracer indexing issues


mnlevy1981:
In some places CISO tracers are indexed from `1:ecosys_ciso_tracer_cnt` and in others they are `ecosys_tracer_cnt+1:ecosys_tracer_cnt+ecosys_ciso_tracer_cnt`; we want to use the latter in all instances and rely on the `ciso_on` flag to determine whether to do CISO computations or not.

mnlevy1981:
Or use `ecosys_ciso_tracer_begin` and `ecosys_ciso_tracer_end`, but the point is that these should always refer to the global MARBL tracer count rather than the CISO slice of tracers.

mnlevy1981:
Addressed in pull request #48

Issue #9, opened 2016-01-15
Apply land_mask to all tracers at once


mnlevy1981:
Currently, `ecosys_driver_init_tracers_and_saved_state()` applies the land mask to ecosys tracers, whie `ecosys_driver_ciso_init_tracers()` applies land mask to ciso tracers. Why not apply land mask all at once after initializing all tracers?

mnlevy1981:
Same for loops setting `column_tracer_module` and `dtracer_module`

mnlevy1981:
Addressed in pull request #48

Issue #4, opened 2016-01-12
Does vflux_flag belong in the driver or in MARBL?


mnlevy1981:
Currentl `vflux_flag` resides in the driver and is set as follows:

```
    vflux_flag(:) = .false.
    vflux_flag(dic_ind) = .true.
    vflux_flag(alk_ind) = .true.
    vflux_flag(dic_alt_co2_ind) = .true.
```

But maybe we just want the driver to tell MARBL if it is using a virtual salt flux and then MARBL can set these flags accordingly? Where does this fall on the science vs infrastructure spectrum?

mnlevy1981:
Addressed in pull request #48

These fields remain in the driver; the corresponding surface averages are now solely in the driver (we moved namelist control of the surface average values to `&ecosys_driver_nml`)

Issue #2, opened 2016-01-12
Should we assume that ecosys_ind_begin = 1?


mnlevy1981:
`ecosys_ind_begin = 1` and `ecosys_ind_end = ecosys_tracer_cnt` because we call `set_tracer_indices` with `'ECOSYS'` before we call it with `'CISO'`. Throughout MARBL we are alternating between declaring vectors with `dimension(ecosys_tracer_cnt)` and `dimension(ecosys_ind_begin:ecosys_ind_end)`. We need to be consistent, but it is unclear which is preferable.

mnlevy1981:
I would argue that, if it is up to the driver to determine `ecosys_ind_begin` and `ecosys_ind_end`, then MARBL either needs to (a) check that the driver does start at 1, or (b) allow for the fact that other implementations maybe number the CISO tracers first.

A third option might be to move the index determination into MARBL.

mnlevy1981:
Addressed in pull request #48

Resolution: `ecosys_base_ind_beg`, `ecosys_base_ind_end`, `ciso_ind_beg`, and `ciso_ind_end` are all part of the `marbl_tracer_index_type`, which is not available to the driver. Instead, `marbl_tracer_metadata_type` contains a new field (`tracer_module_name`) to tell the driver which module a specific tracer belongs to.

PR #48, opened 2016-04-21
Bugfix/tracer index

mnlevy1981:
**Updates on this Branch**

The main goal of this pull request is to remove as much semantic knowledge about which tracer modules MARBL is running (ecosys vs ciso vs both) from the driver as possible. To that end:
1. Internally, tracers are indexed from 1 to N across all modules [previously, indices were 1-27 for ecosys and 1-14 for CISO, now it is 1-41 when using both]
   - these indices were moved into the `marbl_tracer_index_type`
   - tracers associated with autotroph / zooplankton are included in the indexing because the above data type also contains variables (`auto_inds` and `zoo_inds`) of the `marbl_living_tracer_index_type`
2. The total tracer count is determined in MARBL (and the driver passes an expected tracer count to `marbl_instance%init()` to verify), rather than being set in the driver
3. Information about where to read initial data for each tracer is processed entirely in MARBL (before it was up to the driver to realize that `tracer_init_ext` may override defaults)
   - The tracer file metadata (a variable of type `marbl_tracer_read_type` ) is now part of the `marbl_interface_class`
   - The `marbl_tracer_metadata_type` data type now contains a `tracer_module_name` field so that the driver knows if a specific tracer is part of ecosys or ciso.

There was some related code clean-up:
1. Several namelist variables were removed from `&ecosys_nml` and `&ecosys_ciso_nml`
   - MARBL does not know about virtual fluxes, so things like reference surface averages belong in the driver (if the GCM requires them).
   - Specifically, `use_nml_surf_vals`, `surf_avg_*_const`, and `comp_surf_avg` are no longer part of MARBL
2. `non_autotroph_tracer_cnt` and `non_autotroph_ciso_tracer_cnt` are no longer needed

**Testing**

_Versions_
CIME: 4.4.8
CICE: cice5_20160309

_Results_
- yellowstone_intel: aux_pop and aux_pop_MARBL passed (no build- or run-time errors), and baseline comparisons with the same CIME / CICE tags but POP trunk from 20160317 were bit-for-bit. Namelist comparison tests failed in several cases due to moving some variables into `&ecosys_driver_nml`
- hobart_nag: aux_pop_MARBL passed (no build- or run-time errors); no baseline comparisons were done because I was keeping those on scratch and they got wiped. (sidenote: I need a better workflow for testing on hobart)

**Driver Updates**

This will require updates in POP as well; those driver-side changes are in

https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n08_marbl_dev_n03_cesm_pop_2_1_20160407/

Compare against

https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_tags/marbl_dev_n03_cesm_pop_2_1_20160407/

To see changes introduced.


mnlevy1981:
I should mention that this addresses Issue #43

mnlevy1981:
Added bonus: this also addresses Issue #9

mnlevy1981:
Another added bonus: this also addresses Issue #2

mnlevy1981:
And Issue #4

mnlevy1981:
Suggestions from @klindsay28 
1. Rather than putting `marbl_tracer_index_type` in `marbl_interface_types`, put it in `marbl_internal_types` and add a `get_tracer_index` routine to `marbl_interfaces.F90` -- POP can use the `get` routine to request index numbers for virtual fluxes.
2. In the `marbl_tracer_index_type` constructor:
   1. wrap the log message regarding CISO in a `if (ciso_on) then` statement
   2. Data type should contain `marbl_ind_beg` (and end, and the ciso versions), those should not come from `marbl_sizes`, that's a possible threading conflict waiting to happen
3. Rather than a separate clause for `ciso_on`, `marbl_update_tracer_file_metadata` should take `tracer_init_ext` or `ciso_tracer_init_ext` as an argument (will reduce duplicate code)
4. `marbl_diagnostics_mod` is not consistent in variable naming -- use `dtracers` rather than `dtracer` throughout
5. Set `units` and `tracer_module` for all CISO tracers in a loop at beginning of `ciso_init_tracer_metadata`
6. Remove `num_tracers` from `ciso_set_interior` and `ciso_set_surface_forcing`

mnlevy1981:
896065c addresses both parts of (2) in the comment above -- it also introduces more descriptive names for the various tracer counts:

`marbl_total_tracer_cnt` -- total number of tracers in MARBL (was `ecosys_used_tracer_cnt`)
`ecosys_base_tracer_cnt` -- number of tracers in the base ecosys module (was `ecosys_tracer_cnt`)
`ciso_tracer_cnt` -- number of tracers in the CISO module (was `ecosys_ciso_tracer_cnt`)

mnlevy1981:
78cfb53 addresses (5), 765c0f2 addresses (6)

mnlevy1981:
bdb792f addresses (4) above

mnlevy1981:
16fe6cd addresses (3), just leaves the big refactor (moving `marbl_tracer_index_type` from `interface_types` to `internal_types` and adding a `get_index` routine)

mnlevy1981:
Starting aux_pop_MARBL testing with

 https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n11_marbl_dev_n03_cesm_pop_2_1_20160407

Assuming all goes well, I'll start aux_pop tests tonight and merge marbl_dev_levy -> marbl_dev over the weekend. If I can't get to it this weekend, it'll happen first thing Monday morning.

mnlevy1981:
Between aux_pop_MARBL and aux_pop, it looks like all the restart tests with `DEBUG=TRUE` and the CISO tracer module are failing... I don't know if this is an issue in the driver or something in MARBL but I'll investigate over the weekend.

Issue #40, opened 2016-03-25
CECO compsets have runtime error with gnu compiler


mnlevy1981:
When building CESM with the gnu compiler, compsets that use the ecosystem ocean tracer module produce the following runtime error:

```
     (passive_tracer_tools:read_field_3D) reading alk_riv_flux from /glade/p/cesmdata/cseg/inputdata/ocn/pop/gx1v6/forcing/river_nutrients_GNEWS2000_gx1v6.nc
     (passive_tracer_tools:read_field_3D) reading doc_riv_flux from /glade/p/cesmdata/cseg/inputdata/ocn/pop/gx1v6/forcing/river_nutrients_GNEWS2000_gx1v6.nc
 (tavg_requested) id =            0
     (tavg_requested) FATAL ERROR: invalid tavg id
------------------------------------------------------------------------

POP aborting...
 FATAL ERROR: invalid tavg id

------------------------------------------------------------------------
```

The error does not crop up when using the Intel or NAG compilers.

klindsay28:
This appears to be a problem in one of the source files comprising the POP driver for MARBL, not in MARBL itself. The patch below, to POP, enables CECO and GECO compsets to run with gnu. However, even with this patch, tests with ciso enabled are failing with gnu.

# Index: ecosys_tavg.F90

--- ecosys_tavg.F90 (revision 78054)
+++ ecosys_tavg.F90 (working copy)
@@ -235,7 +235,7 @@

```
 nblocks_clinic = size(flux_diags,4)
```
-    !$OMP PARALLEL DO PRIVATE(iblock)
-    !$OMP PARALLEL DO PRIVATE(iblock,i)
   do iblock=1,nblocks_clinic
  
  ```
  associate (diag_cnt => marbl_instances(iblock)%surface_forcing_diags%diag_cnt)
  ```

klindsay28:
Whoops, I didn't preview, so I didn't see what github was going to do to cut-n-paste of the patch. My bad. I've attached a .txt file of the patch this time. The fix is included as a patch because the code in question is outside of git.
[patch.txt](https://github.com/NCAR/MARBL/files/233275/patch.txt)

mnlevy1981:
This fix is now in both the POP trunk and the `marbl_dev` branch of POP; additionally, an issue that prevented running CISO with GNU or NAG has been fixed via pull request #54

PR #54, opened 2016-04-28
bugfix: eliminate runtime error in ciso diagnostics

klindsay28:
remove store to unallocated field_3d components for vertical integrals
remove un-necessary store of zero to field_2d components for vertical integrals

passes aux_pop_MARBL test suite with gnu, with pop2/trunk_tags/cesm_pop_2_1_20160426

### DIFF
--- a/src/marbl_diagnostics_mod.F90
+++ b/src/marbl_diagnostics_mod.F90
@@ -4089,7 +4089,6 @@ contains
 
     ! Vertical integrals - CISO_Jint_13Ctot and Jint_100m_13Ctot
 
-    diags(ind%CISO_Jint_13Ctot)%field_3d(:, 1) = c0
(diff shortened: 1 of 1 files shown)


mnlevy1981:
Huh, I don't know how I got credit for adding the "In Progress" label... I tried to set up

https://waffle.io/NCAR/MARBL

And maybe that changed something?

Anyway, this addresses #40 and I'll bring it to master after resolving pull request #48

mnlevy1981:
2 things:

1) I think the initialization in line 3469 is un-necessary as well
2) This fix is necessary (and sufficient) to run with NAG so GNU is not the only compiler that flagged this bug. I think we should add a CISO test for hobart_nag to the aux_pop_MARBL test suite as part of this pull request.

mnlevy1981:
@I just realized that (2) above is actually in the POP tag, not MARBL... so it's just a one-line fix (removing line 3469 from `marbl_diagnostics_mod.F90`) in MARBL.

PR #55, opened 2016-04-29
Enhancement/more compilers

mnlevy1981:
This branch adds stand-alone build support for the Cray compiler (tested on edison) and the NAG compiler (tested on hobart). I had started a separate branch for NAG work, and may want to restructure the Makefile based on commit d6353f1(I like the organization of the targets better in that commit) before accepting this pull request.


mnlevy1981:
I was able to add Cray support and clean up the Makefiles in the "non-namelist initialization" work, so this pull request is not necessary.

PR #61, opened 2016-05-10
Science/dust flux driver

klindsay28:
corresponding POP marbl_dev_klindsay tag is
marbl_dev_klindsay_n06_marbl_dev_n07_cesm_pop_2_1_20160426


klindsay28:
aux_pop_MARBL test suite on hobart with nag passes with following execptions
   SMS_Ld_D test fails
   some nlcomp tests fail because added namelist variables
(the baseline is to the previous tags and commits on the respective branches, not to NCAR/master

mnlevy1981:
The code looks good, but should the following be addressed before the pull request is accepted?

```
! FIXME : check coefficient values in formula here
```

https://github.com/NCAR/MARBL/pull/61/files?w=1#diff-a1264867a6cdb3406ca4dadf7e51bfeaR2986

I still need to look at changes to `ecosys_driver.F90` in the POP repository.

mnlevy1981:
Another question that came up when I realized you were adding new MARBL namelist variables (i.e. while looking at the driver code): you initialize `iron_frac_in_dust` with

```
iron_frac_in_dust            = 0.035_r8 * 0.01_r8
```

https://github.com/NCAR/MARBL/pull/61/files?w=1#diff-a1264867a6cdb3406ca4dadf7e51bfeaR498

Is there a reason you do this multiplication explicitly? From the comments surrounding the definition of the variable, it seems like it is unitless... so it's not a unit conversion factor (as far as I can tell). Initializing with

```
iron_frac_in_dust            = 3.5e-4_r8
```

seems "cleaner" to me, unless there is a good reason to define it as is.

My only other comment from the driver (at the moment) is that you have another FIXME that should be addressed before merging (`ecosys_driver.F90`, line 1664):

```
! FIXME : is this scaling correct if dust_flux_source = 'driver' ?
```

PR #64, opened 2016-05-17
Enhancement/generalize saved state

mnlevy1981:
This branch generalizes the saved state datatype. It requires

https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n19_marbl_dev_n10_cesm_pop_2_1_20160513

and passes all aux_pop_MARBL tests (yellowstone_intel, yellowstone_gnu, and hobart_nag).

@klindsay28 I'm hoping we can talk about this on Friday or Monday.

### DIFF
--- a/src/marbl_diagnostics_mod.F90
+++ b/src/marbl_diagnostics_mod.F90
@@ -4185,12 +4185,14 @@ contains
        surface_tracer_fluxes,                       &
        marbl_tracer_indices,                        &
        saved_state,                                 &
+       saved_state_ind,                             &
        surface_forcing_output,                      &
        surface_forcing_diags)
 
     ! !DESCRIPTION:
     !  Compute surface fluxes for ecosys tracer module.
 
+    use marbl_interface_types , only : marbl_surface_saved_state_indexing_type
     use marbl_share_mod       , only : lflux_gas_o2
     use marbl_share_mod       , only : lflux_gas_co2
     use marbl_share_mod       , only : iron_flux_file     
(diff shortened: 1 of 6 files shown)


mnlevy1981:
Looking through the code diffs after commit fbb3be2, there is still a little more work to be done. I'll continue to clean this code up and let you know when it's actually ready to be reviewed.

mnlevy1981:
I'm in the process of setting up a `marbl_saved_state_mod`, and I've noticed a slight discrepancy in how the code is organized. `marbl_log_type` is defined in `marbl_logging.F90`, despite being part of the interface - I would have expected this datatype to be in `marbl_interface_types.F90`. Not sure what the best thing to do is, but it's something to pay attention to when we reach the code-cleanup stage of the project.

mnlevy1981:
I think this branch is ready for review. I probably need better comments in `marbl_saved_state_mod.F90` but I like how the code itself looks - the generalized saved_state type initializes in a similar manner to the surface forcing output type: it allocates memory for each `state%add()` call, rather than requiring a hard count of the number of saved state fields or cycling through the fields twice (once to count, once to actually set up the data type).

mnlevy1981:
Forgot to note that a2baa5e passes `aux_pop_MARBL` and is bit-for-bit with latest trunk on yellowstone (intel and gnu) as well as hobart (nag)

mnlevy1981:
9274328 using

https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n20_marbl_dev_n10_cesm_pop_2_1_20160513

passes all tests and is bit-for-bit with baselines (`aux_pop` on `yellowstone_intel`, `aux_pop_marbl` on `yellowstone_intel`, `yellowstone_gnu`, and `hobart_nag`)

PR #65, opened 2016-05-17
Add one more status_log call

mnlevy1981:
One routine in marbl_diagnostics_mod (linear_root()) was still using print\* to
write to stdout; added a status_log argument to the interface and passed the
message out.

### DIFF
--- a/src/marbl_diagnostics_mod.F90
+++ b/src/marbl_diagnostics_mod.F90
@@ -4124,8 +4124,13 @@ contains
       return
     end if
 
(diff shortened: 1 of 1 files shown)


mnlevy1981:
This is just one more case that should've been fixed when we closed #23 (pull request #38)

mnlevy1981:
Passes aux_pop_MARBL tests on yellowstone (intel and gnu), and builds successfully with nag on hobart. No change needed in the driver.

PR #71, opened 2016-06-05
Tag Summary: implement global averages, scalar running means, ladjust…

klindsay28:
…_bury_coeff

global averages and scalar running means are computed in the GCM
   running means will be moved to MARBL later
introduce set_global_scalars to interface for consistent setting of time-varying scalars in MARBL
introduce ladjust_bury_coeff
   If true, then model tracks running mean of C, P, and Si input and burial
   averages and adjusts bury coefficients to balance losses against inputs
   on 10 year timescale. Use Newton's Method on running means to solve for
   balancing burial coefficient.

Testing:
with marbl_dev_klindsay_n09

aux_pop_MARBL (intel and gnu): (baseline comparison to n08 tag)
   all tests pass, with the following exceptions
      some nlcomp tests fail because of added namelist variables
      cice_ecosys_ladjust_bury_coeff fails baseline compare because it is a new test

aux_pop_MARBL on hobart (nag) will be done after commit
   lack of subsequent tags indicates success

Files Modified:
    modified:   src/marbl_ciso_mod.F90
    modified:   src/marbl_interface.F90
    modified:   src/marbl_interface_types.F90
    modified:   src/marbl_internal_types.F90
    modified:   src/marbl_mod.F90
    modified:   src/marbl_parms.F90

### DIFF
--- a/src/marbl_ciso_mod.F90
+++ b/src/marbl_ciso_mod.F90
@@ -1494,7 +1494,6 @@ contains
     !----------------------------------------------------------------------------------------
     
     use marbl_parms     , only : denitrif_C_N
-    use marbl_parms     , only : parm_POMbury
     use marbl_parms     , only : spd
 
     implicit none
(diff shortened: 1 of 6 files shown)


klindsay28:
The latest commit, together with marbl_dev_klindsay_n11_marbl_dev_n11_cesm_pop_2_1_20160513, passes aux_pop_MARBL on hobart/nag (except for SMS_Ld2_D.T62_g37.CECO, which is an expected fail). So this pull request is ready to go.

Issue #73, opened 2016-06-07
MARBL must return tracer count from initialization


mnlevy1981:
This is mentioned in the MARBL design document -- currently the MARBL initialization wants `gcm_tracer_cnt` as an `intent(in)` and the consistency check (is MARBL running with the right number of tracers?) is done in MARBL. It makes more sense to have MARBL pass its tracer count as an `intent(out)` and the GCM can make sure it has allocated enough space to handle them all.

I will fix this the branch where I am overhauling the initialization routine / adding non-namelist options for setting parameters.

mnlevy1981:
This has been implemented in a branch and will be fixed when #80 is merged onto master.

Issue #70, opened 2016-06-02
Separate initialization into config and init


mnlevy1981:
Before calling `marbl%init()` to initialize parameters (as defined in the [design document](https://www.overleaf.com/5152744hjbryy#/16083348/)), we want to call `marbl%config` to set the general MARBL configuration. There should be a `marbl_config_type` that is similar to the `marbl_parms_type` that I am working on in my enhancements/initialization branch, so variables in `marbl_config` can be set via namelist (new `&marbl_config_nml`) or by puts and gets called from the driver.

Three notes from a discussion with @klindsay28 and @matt-long 
1. Memory for objects like the autotroph and zooplankton derived type should be allocated at the very beginning of init
2. This will let us make the namelist string optional in both config and init: MARBL will set default values in both routines, but only try to read the namelist if it is present. Otherwise users will need to use put calls from the driver to change values.
3. Configuration variables will be set in `marbl%config` and will be locked in the call to `marbl%init` - so any `marbl_config_put` calls after `marbl%init` will abort

mnlevy1981:
This has been implemented in a branch and will be fixed when #80 is merged onto master.

Issue #67, opened 2016-05-19
use of pi in marbl_ciso_mod.F90


mnlevy1981:
`pi` is a module variable in `marbl_ciso_mod.F90` (that is set every time step, per #36), but for some reason `marbl_ciso_mod:fract_keller_morel` declares a local copy of `pi` without initializing it. I think the fix is as simple as removing the local declaration so the routine uses the module variable.

mnlevy1981:
Fixed in  d5adaf8 (currently still on one of my branches, will close issue when pulled to master)

Issue #62, opened 2016-05-11
Better interface between MARBL parameters and GCM driver


mnlevy1981:
This will address part of #11 and all of #51
1. `marbl_parms_mod.F90` will be home to all MARBL namelist variables (expanding definition of "parameters" in design document to mean anything set via namelist, differentiated from "configuration" which is set by the complexity level / tracer modules you want from MARBL: `ciso_on` is a configuration setting, as are autotroph and zooplankton counts. [Gray area: actual autotrophs and zooplankton used in the run?]
2. MARBL will use `marbl_parms_define()` to set up meta-data: `sname`, `lname`, `units`, `type`
3. We will provide the following routines to access / set parameter values:
   1. `marbl_parms_put()`: takes short name (or index?) and value as arguments, sets appropriate parameter to desired value
   2. `marbl_parms_get_metadata()`: either take index as intent(in) and return any (or all) metadata via optional arguments, or take short name as intent(in) and return rest of metadata
   3. `marbl_parms_get_val()`: take index or short name as intent(in), return value of parameter as intent(out)

@klindsay28 wants the GCM driver to be able to run the following:

```
do n=1,marbl_parms_cnt()
  call marbl_parms_get_metadata(n,sname,type)
    select case (trim(type))
      case('r8')
        call marbl_parms_get_val(n, val=r8val)
        print*, trim(sname), " = ", r8val 
      case('i4')
        call marbl_parms_get_val(n, val=i4val)
        print*, trim(sname), " = ", i4val 
    end select
end do
```

I think we want a similar routine available in MARBL as `marbl_parms_write_to_log()` that will  use MARBL's `marbl_log_type` to provide the entire list in a pre-determined format (to be clear, I want this in addition to the driver code Keith requested - I think it would be good to have both)

NOTE: we want the driver to use these routines from `marbl_interface`, so we need something like

```
module marbl_interface

  use marbl_parms_mod, only : marbl_parms_get_val

  public :: marbl_parms_get_val
```

Goal: have API finished by May 25 and then we can talk about a `marbl_parms_type` to store all the parameters or other under-the-covers changes that shouldn't affect the driver.

mnlevy1981:
I think this ticket will actually address all of #11 because I'll be touching several different modules. Here is what MARBL looks like as of v0.4.0:
- `marbl_ciso_mod.F90`: CISO tracer module
- `marbl_co2calc_mod.F90`: compute co2calc for surface, also various co3 terms
[...]

PR #80, opened 2016-06-22
More robust initialization

mnlevy1981:
This branch addresses several issues with setting up MARBL to run:
1. All parameters are written to the log file now, not just those specified by the namelist of the run (#16)
2. The namelist is optional, parameters can be changed via `put()` calls (#62)
3. Tracer count is an `intent(out)` from MARBL; GCM is responsible for consistency check (#73)
4. Some parameters were specified in multiple modules (#11, #36, #67)
5. Setup has been split into two phases (`config`, which nominally is used to set variables that affect tracer count and / or are not parameters; and `init`, which is used to set parameter values that do not affect tracer count). There is also a `complete_config_and_init` phase. (#70)

Additionally, this branch introduces a stand-alone driver. Currently there is just a single test that can be run, which runs through the entire setup process and then prints the log to stdout, but it will hopefully be easy to add additional tests as the need arises.


mnlevy1981:
There is a decent amount of code clean-up to do before merging this branch. I'll update this pull request as I recognize them, but one that stood out while I was resolving conflicts (this was branched off marbl0.5.1 and I brought marbl0.6.0 in yesterday) is that `marbl_set_rmean_init_vals` should be called from `complete_config_and_init` rather than `init`.

mnlevy1981:
Another task to do: address #57 -- GCM needs to explicitly tell MARBL it can handle global sums, otherwise MARBL should return an error if it is configured to request a global mean (currently, I think that just means `ladjust_bury_coeff = .true.`). It might also be worthwhile to move all the global sums / means code into a new `marbl_global_ops_mod.F90`, which would also contain the `lgcm_has_global_ops` flag - I won't do that without some discussion beforehand, though.

mnlevy1981:
This branch has all of the functionality I intend to add, so I believe it's ready for review. It's a very large merge, so I'll be in touch with @klindsay28 to figure out the best way to break up the code review... it probably makes the most sense to look at the infrastructure changes / additions first: I updated `bld-test/Makefile` and `src/makedep.py` plus created the `driver/` and `test/` directories. Once that's understood, we can roll up our sleeves and look at the Fortran changes.

mnlevy1981:
Using this branch of MARBL with

 https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n25_marbl_dev_n12_cesm_pop_2_1_20160513

passes `aux_pop_MARBL` tests on yellowstone (intel and gnu) and hobart (nag)

Issue #75, opened 2016-06-10
redesign specification of tracer initialization


mnlevy1981:
Specification of tracer initialization should be up to the driver rather than MARBL

mnlevy1981:
Available in the marbl0.9.0 tag

Issue #56, opened 2016-05-02
redesign specification of forcing fields


klindsay28:
have MARBL specify, in a generic sense, which fields it needs,
but specification of forcing field source resides in driver

mnlevy1981:
Note that there are forcing fields for CISO that have not been moved to the `marbl_forcing_fields_type` yet (atmospheric D13C and D14C); as part of resolving this ticket, we need to handle these fields consistently as well.

mnlevy1981:
Re: the comment from June 2 about needing to move D13C and D14C into `marbl_forcing_fields_type` -- those two fields are now in the appropriate place.

@klindsay28 and @matt-long -- can you guys verify that this is an accurate representation of where things stand for forcing fields and also a reasonable description of what we want the code to look like after completing the work in this ticket? I think I might be missing something, maybe other places in the code that use the `marbl_forcing_fields_type` besides the surface forcing? Or maybe a place where we _could_ use a more generalized `marbl_forcing_fields_type` but can't use the current implementation?

---

Currently, surface forcing is handled in MARBL in the following manner:
1. In `config` (either via `marbl_forcing_tmp_nml` or `put()` commands), MARBL determines some settings such as "what file will we read `DIN_RIV_FLUX` from?" and "will we get ice fraction and piston velocity from the driver or will they be read from a file?"
2. In `init`, MARBL sets the internal index for where in the forcing fields datatype the following fields will be stored (indices of 0 imply the field is not requested by MARBL), as well as whether the field will be read from a file or received from the driver
   1. `surface_mask`
   2. `u10_sqr`
   3. `ifrac`
   4. `sst`
   5. `sss`
   6. `atm_pressure`
   7. `xco2`
   8. `xco2_alt`
   9. `xkw`
   10. `dust_flux`
   11. `black_carbon_flux`
   12. `iron_flux`
   13. `nox_flux`
   14. `nhy_flux`
   15. `din_riv_flux`
   16. `dip_riv_flux`
   17. `don_riv_flux`
   18. `dop_riv_flux`
   19. `dsi_riv_flux`
   20. `dfe_riv_flux`
   21. `dic_riv_flux`
   22. `alk_riv_flux`
   23. `doc_riv_flux`
   24. `d13c`
   25. `d14c`
   26. `d14c_glo_avg`
3. During the time stepping loop, the driver populates `marbl%surface_input_forcings` as directed (either by reading the file or just dropping in the correct variable)

---

What we want:
[...]

PR #94, opened 2016-09-21
Enhancement/improve tracer and forcing spec

mnlevy1981:
Per issues #56 and #75, information about (a) how to initialize tracers and (b) where surface forcing fields are coming from (file, driver, namelist constant, etc) has been moved from MARBL back to the driver layer. Interior tracer restoring initialization ("what file contains the climatology?") has also been moved to the driver. So the driver needs to check three things:
1. `marbl_instance%tracer_metadata` -- contains the names / units of all the tracers MARBL is running with (driver needs to populate `marbl_instance%column_tracers` and this is how it knows which tracer is which; `column_tracers` dimensions are `(nt, km)`)
2. `marbl_instance%surface_forcing_metadata` -- contains the names / units of all forcing fields that MARBL requires
3. `marbl_instance%interior_forcing_input%tracer_names` -- contains the names of all the tracers that MARBL will restore towards a climatology

Right now there is no mechanism in place to ensure the driver has checked these two lists and will be providing all the requested data. Should there be?


mnlevy1981:
As of 34906ac, I think this branch is ready to merge back onto `master` (it contains all the minor tweaks @klindsay28 and I talked about during the code review). I'll wait until the POP driver updates are done as well before accepting this pull request, though, just as one more interface test.

mnlevy1981:
The URL to see the changes made since the code review is

https://github.com/mnlevy1981/MARBL/compare/e7a1dff...mnlevy1981:34906ac

Issue #114, opened 2016-11-30
need some safety checks in clean target for stand-alone driver Makefile


mnlevy1981:
`tests/driver_src/Makefile` uses the `realpath` function to define where to put object and module files, as well as `libmarbl.a` and the MARBL executable:

```
OBJ_ROOT = $(realpath ../obj)
OBJ_DIR = $(OBJ_ROOT)/$(COMP_NAME)/driver
INC_DIR = $(realpath ../../include)
LIB_DIR = $(realpath ../../lib)
EXE_DIR = $(realpath ../driver_exe)
SRC_DIR = $(realpath ../../src)
DRV_DIR = $(realpath .)
```

The clean target removes files in these directories:

```
clean_exe:
        rm -f $(EXE)

clean: clean_exe
        rm -f $(OBJ_ROOT)/*/*.* $(OBJ_ROOT)/*/*/* $(LIB_DIR)/* $(INC_DIR)/*/*
```

If a directory does not exist, then realpath returns an empty string, resulting in

`rm -f /*/*.* /*/*/* /* /*/*`

not a very good command to run. This could happen if, for example, you think the following would be an easy way to clean up any lingering build files (WARNING: don't try this at home) (WARNING: or at work):

```
$ cd $MARBL_DIR
$ make -f tests/driver_src/Makefile clean
```
An easy patch would be to update the clean target:

```
clean: clean_exe
        rm -f $(OBJ_ROOT)/*/*.{o,mod,d} $(OBJ_ROOT)/*/driver/*.{o,mod,d} $(LIB_DIR)/*.a $(INC_DIR)/*/*.mod
```
A better option (and the two are not mutually exclusive) would be to check if `$OBJ_ROOT`, `$LIB_DIR`,`$INC_DIR`, or `$EXE_DIR` are empty and produce an error instead of continuing with the build in that case.

mnlevy1981:
The Makefile in `$MARBL_DIR/src` does not use the `realpath` function, and therefore doesn't pose the same issue. Really, though, that's an oversight on my part -- I try to keep the two Makefiles as similar as possible, and will probably introduce `realpath` (and whatever fixes come out of this ticket) when updating the driver Makefile.

mnlevy1981:
fixed in marbl0.14.0 release

PR #115, opened 2016-11-30
Bugfix/makefile cleanup

mnlevy1981:
Fix issue #114 -- safety checks in the two Makefiles before continuing to build (or clean) with specified directories.


mnlevy1981:
Testing:

1) built (and did a clean build) on both my local laptop and hobart
2) after clean build, verified that the object, module, etc files were in fact removed (I updated the clean target)
3) ran `make clean` from the non-source directory and verified that errors out rather than cleaning with wrong `OBJ_DIR` and such

```
$ make -f tests/driver_src/Makefile clean
tests/driver_src/Makefile:16: *** Can not find $OBJ_ROOT.  Stop.
$ make -f src/Makefile clean
src/Makefile:12: *** Can not find $OBJ_DIR.  Stop.
```

I would have preferred to have more detail in the error message, but the Makefile language would have made that extremely clunky. Also, I will test on yellowstone when the machine is available again and update this ticket with what I assume will be good news.

mnlevy1981:
Have tested that this branch builds (and cleans) as expected on hobart and yellowstone, as well as my local machine. Keith has some commits that will change the Makefile (increasing tracer count), so I'll wait until we've merged those before having a code review and actually accepting this pull request. The best summary comes from the 4f1e252 commit message:

[W]hereas previously undefined values of OBJ_DIR, etc would lead to

```
rm -f /*/*.* /*/*/* /* /*/*
```

we now have safety checks to error out, but even if those are not triggered the worst case call would be

```
rm -f /{gnu,intel,nag,pgi,cray}/marbl_*.{o,mod,d} /{gnu,intel,nag,pgi,cray}/driver/marbl_*.{o,mod,d} /{gnu,intel,nag,pgi,cray}/driver/marbl.{o,mod,d} /{gnu,intel,nag,pgi,cray}/marbl_*.mod /libmarbl*.a
```

mnlevy1981:

PR #117, opened 2016-12-03
Enhancement/interior forcing spec

mnlevy1981:
I've generalized the specification of the interior forcing fields (including tracer restoring) in MARBL; at this point in time, I have not made any corresponding changes to POP so this pull request is not ready to be accepted... but I'd like to go over the MARBL changes while I work on the POP branch, in case some of the interface decisions I've made need to be changed.

I'll add a comment to this pull request once POP tests are complete. The stand-alone driver builds and runs on yellowstone (intel and gnu), hobart (nag, intel, and gnu), and my local machine (gnu); the output for the `request_forcings` test (using the `marbl_in_with_restore` namelist) is

```
Requested surface forcing fields
---
1. surface_mask
2. u10_sqr
3. sss
4. sst
5. Ice Fraction
6. Dust Flux
7. Iron Flux
8. NOx Flux
9. NHy Flux
10. DIN River Flux
11. DIP River Flux
12. DON River Flux
13. DOP River Flux
14. DSi River Flux
15. DFe River Flux
16. DIC River Flux
17. ALK River Flux
18. DOC River Flux
19. Atmospheric Pressure
20. xco2
21. xco2_alt_co2
22. d13c
23. d14c
24. d14c_gloavg

Requested interior forcing fields
---
1. Dust Flux
2. PAR Column Fraction
3. Surface Shortwave
4. Temperature
5. Salinity
6. Pressure
7. Iron Sediment Flux
8. PO4 Restoring
9. NO3 Restoring
10. SiO3 Restoring
```

The `request_restore` test currently does not work; given the appearance of restoring in the above output, maybe it's no longer needed? Alternatively, I need to rework the logic to list the requested tracers for restoring.

### DIFF
--- a/src/marbl_ciso_mod.F90
+++ b/src/marbl_ciso_mod.F90
@@ -38,7 +38,6 @@ module marbl_ciso_mod
   use marbl_interface_types , only : marbl_tracer_metadata_type
   use marbl_interface_types , only : marbl_diagnostics_type
   use marbl_interface_types , only : marbl_domain_type
-  use marbl_interface_types , only : marbl_interior_forcing_input_type
 
   use marbl_internal_types  , only : autotroph_parms_type
   use marbl_internal_types  , only : column_sinking_particle_type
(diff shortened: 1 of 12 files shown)


mnlevy1981:
Following a code review with @klindsay28 we decided that it made more sense to put the forcing fields coming from the GCM in a datatype rather than a flat array (the flat array worked well for the surface forcings, but some interior forcings are scalar while others are arrays... and some of the arrays are different dimensions (`temperature` is length `km` while `PAR Column Fraction` is length `num_PAR_subcols`). So originally I had introduced two flat arrays for `interior_input_forcings_0d` and `interior_input_forcings_1d` but the new `marbl_forcing_fields_type` is a better fit. For symmetry in the code, I also converted the surface forcings to this new data type.

PR #120, opened 2016-12-10
Tag Summary: subtract tracer restoring terms from Jint integrals

### LINKED ISSUE #118: Jint integrals should subtract out tracer restoring terms
The Jint integrals are intended to compute vertical integrals of elemental (C, P, N, Si) tendencies, summing elemental tendencies over tracers and subtracting out non-conservative tendency terms. The idea is that the result should be within round-off of zero. Tracer restoring terms, which are non-conservative, are currently not subtracted out, and they should be.

Note that this bug only impacts a diagnostic output, not the tracer interior tendencies or surface fluxes.

klindsay28:
fixes #118

Testing:
with cesm2_0_beta04 and marbl_dev_n24

aux_pop_MARBL yellowstone/intel and yellowstone/gnu: (baseline comparison to marbl_dev_n23 tag+PAR_threshold)
   all tests pass, except for expected failure of ERI and baseline failure
      in ecosys_restore because Jint integrals change

Files Modified:
	modified:   marbl_diagnostics_mod.F90
	modified:   marbl_mod.F90

### DIFF
--- a/src/marbl_diagnostics_mod.F90
+++ b/src/marbl_diagnostics_mod.F90
@@ -4090,6 +4090,7 @@ contains
        sed_denitrif, other_remin, nitrif, denitrif,   &
        column_o2, o2_production, o2_consumption,      &
        fe_scavenge, fe_scavenge_rate,                 &
+       interior_restore,                              &
        marbl_interior_forcing_diags,                  &
        marbl_status_log)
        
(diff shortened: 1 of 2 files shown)


mnlevy1981:
I was thinking we should bring this onto master today as well, except it looks like there are conflicts that need to be resolved...

mnlevy1981:
Actually, I think we were waiting on this as it will be answer changing (for the `ecosys_restore` test, at least). Sorry, still working on remembering everything I forgot over winter break...

klindsay28:
This change is now on the science/iron_ligand_varying_P_C branch of https://github.com/klindsay28/MARBL with commit a9d4fbff856fcad28199bd6e37433532b0635011, so I'm closing the pull request.

PR #122, opened 2017-01-03
Fix issue 121

mnlevy1981:
Be more specific in what we match in forcing field name when determining restoring fields.


mnlevy1981:
@klindsay28 -- I'd like to bring this to master today, does that sound okay to you?

klindsay28:
Sounds fine to me.

Issue #27, opened 2016-03-21
MARBL timers?


mnlevy1981:
There is a FIXME comment in `marbl_mod` that reads:

```
new marbl timers need to be implemented to turn
on timers here around this subroutine call
```

(subroutine call in question is `marbl_compute_carbonate_chemistry`). Do we want MARBL to be able to provide timing information? If so, how do we incorporate that?

mnlevy1981:
Jim Edwards suggests creating a timing module (marbl_timing_mod.F90) and using `#ifdef` wrappers to support GPTL if requested at build time (otherwise we can fall back to walltime or something). This is already done (in a way) in POP & CIME; POP uses CCSMCOUPLED in `source/timers.F90` to trigger calls to `cime/share/timing/perf_mod.F90` which is just a wrapper for GPTL. My first thought is to cut out the middle man and call GPTL directly if it is available.

This is not currently a high priority, but I would like to finish it sometime this summer... I believe it'll be useful to have #16 finished, because that bugfix will also provide the framework for a stand-alone driver and we'll be able to do some initial testing outside of CESM before trying to tie the MARBL timers in to POP.

mnlevy1981:
I might be falling victim to scope creep, but I am envisioning three possible timers:

1. GPTL
2. MPI timers (based on `MPI_Wtime`)
3. Fortran intrinsic timers (my copy of Metcalf, Reid, and Cohen is in the office, so I need to see if `cpu_time()` is a Fortran intrinsic or just a special function in `gfortran`; assuming the former, it will be the easiest to implement)

This means the test suite needs to handle two additional flags, one if we want to use GPTL and another if MPI is available. My current thought for that is taking two more arguments when running the test suite: `--gptl_loc [LOCATION OF GPTL LIBRARY]` and `--mpi_tasks [# of tasks]`. The defaults would be "no `--gptl_loc` implies run without GPTL" and "no `--mpi_tasks` implies `mpi_tasks`=0  and only use mpirun if `mpi_tasks` > 0". Of course, this opens up a new can of worms because different machines have different `mpirun` commands (`mpirun` vs `mpiexec` vs `mpirun.lsf` vs ???); I'll get this working on my laptop with mpich and then think about how to make it more extensible.

[...]

PR #129, opened 2017-01-30
Add timers to MARBL

mnlevy1981:
MARBL timers use `cpu_time()` by default, but building with `-DMARBL_TIMING_OPT=MPI` will use `MPI_Wtime()` instead and `-DMARBL_TIMING_OPT=CIME` will use the [CIME](https://github.com/ESMCI/cime) wrappers to the GPTL library. There is no support for using the GPTL library without the CIME wrappers.

`marbl_timing_mod.F90` handles most of the heavy lifting. Currently there are four operational timers:

1. the entire 3-stage initialization process
1. `set_surface_forcing()`
1. `set_interior_forcing()`
1. `compute_carbonate_chemistry()` (which is called from within the interior forcing routine).

This addresses Issue #27


mnlevy1981:
There are two fairly big infrastructure changes in this pull request:

1. I added MPI support to the stand-alone driver to test the `MPI_Wtime()` timers, which resulted in some fairly big changes:
    * Introduced `tests/driver_src/marbl_mpi_mod.F90` to provide wrapper for mpi calls
    * Build `marbl.exe` as usual for non-MPI, but build separate `marbl-mpi.exe` for parallel runs
    * More object directories, etc for the build process (so you can switch between parallel and serial tests without rebuilding)
1. Python testing library now uses a class for the testing; this removes the need for global variables

mnlevy1981:
This is kind of on the border of "cleaner code" and "object oriented code for the sake of object oriented code", but I think I should move `marbl_timing_setup_timers()` and `marbl_timing_copy_timing_data()` into the `marbl_internal_timers_type` class.

Issue #24, opened 2016-03-21
MARBL does not need both dz and delta_z


mnlevy1981:
We should remove `dz` from `marbl_domain_type` and all layer depth references should use `delta_z` (POP has a 1D array for using the same levels throughout the model and a 3D array for partial bottom cells; MARBL does not need to differentiate between these two.)

mnlevy1981:
Additionally, does MARBL need `zw`? One option would be to remove it (using `zt` for depth thresholds and whatnot), another option would be to keep `zw` but compute `dz` on the fly instead of requiring all three variables for initialization.

mnlevy1981:
fixed in marbl0.14.0 release

PR #140, opened 2017-02-28
Remove "dz" variable in favor of delta_z

mnlevy1981:
This addresses issue #24, that the domain type does not need both dz and
delta_z. For the fix, I changed all instances of dz on the interface to
delta_z -- this is in marbl_domain_type and also the calls to init().

Note that there will be a small in POP as well (the call to `marbl_instances(iblock)%init()` requires `gcm_delta_z` rather than `gcm_dz`); I need to add that change to marbl_dev_levy before accepting this pull request.


mnlevy1981:

PR #141, opened 2017-02-28
Remove locmip_k1_k2_bug_fix from config

mnlevy1981:
MARBL should not be run with this variable set to `.false.` so I removed the
variable and changed the code to only follow the `.true.` path.

Note that this update also removes the `k1_k2_pH_tot` variable from
`marbl_comp_co3_coeffs()` as that variable is also always set to .true.

There is a required fix in POP as well (removing `locmip_k1_k2_bug_fix` from `&marbl_config_nml` in the `build-namelist` tool).

@klindsay28 -- did you already include a fix for this on your science branch? If so, this pull request can be disregarded.


klindsay28:
I do not have this on my branch.

mnlevy1981:

Issue #112, opened 2016-11-29
revamp how parameters are written to the log


klindsay28:
Write parameters in blocks based on parameter category (see issue #111),
with white space between blocks.

This will lead to add_newline being unnecessary.

mnlevy1981:
fixed in marbl0.14.0 release

Issue #111, opened 2016-11-29
add category field to marbl_single_config_or_parm_type


klindsay28:
this will help organize the structure of parameters,
including how they are written out to the log

mnlevy1981:
fixed in marbl0.14.0 release

PR #142, opened 2017-02-28
Enhancement/improve parm output

mnlevy1981:
Addresses issues #111 and #112 

I'd like some input on what categories we want, but I put in a strawman set that keeps the output in the log readable.

Also, I don't know if we want the "group" element in `marbl_single_config_or_parm_type` since we now assume that we have one instance of `marbl_config_and_parms_type` per namelist (we only have two namelists, `&marbl_config_nml` and `&marbl_parms_nml`)

### DIFF
--- a/src/marbl_config_mod.F90
+++ b/src/marbl_config_mod.F90
@@ -278,12 +278,13 @@ contains
     ! marbl_config_nml !
     !------------------!
 
+    category  = 'config flags'
+
     sname     = 'ciso_on'
     lname     = 'Control whether CISO tracer module is active'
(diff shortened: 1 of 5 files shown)


mnlevy1981:
As part of cleaning up the log output, I introduced a new method to `marbl_logging_type` -- the `log_header` routine takes a string (e.g. `HEADER TEXT`) and writes the following to the log (including whitespace)

```

-----------
HEADER TEXT
-----------

```

(The dashed lines are always the same length as the text it is wrapping)

mnlevy1981:

PR #143, opened 2017-02-28
Make diagnostics array reallocatable

mnlevy1981:
Still some code clean-up to do, but on my laptop the initialization timer reported an increase from ~0.015 seconds to ~0.040 due to additional memory management. I think that's a pretty cheap cost for standardizing how the `%add()` routine works across many different datatypes -- I'm pretty sure diagnostics was the only case where we were only allocating memory once and therefore counting all the `%add()` calls before actually making them.


mnlevy1981:

PR #144, opened 2017-03-01
Bugfixes/combine pull requests

mnlevy1981:
This combines #115, #142, #143, #140, and #141 into a single branch / pull request. The first two pull requests are MARBL-only; the last three require updates in POP as well. I am currently running `aux_pop_MARBL` on hobart and yellowstone; I expect these results to all be bit-for-bit with `marbl_dev_n33`.

I'm going to keep the individual pull requests open for now, as it will make code review a lot easier. After the code review, I will close those pull requests -- this request will stay open and not get merged until after we get @klindsay28 science branch onto master.


mnlevy1981:
all hobart tests passed (except the expected DEBUG / nag failure and a couple of MEMLEAK messages)

mnlevy1981:
all yellowstone tests passed except for a couple of MEMLEAK and TPUTCOMP results)

Issue #99, opened 2016-10-28
Update how grid location of MARBL error is reported to GCM


mnlevy1981:
Currently `marbl_set_surface_forcing` is called for multiple grid cells ("elements") simultaneously, and if MARBL errors out it reports which grid cell in the error log... but there is no way for the driver to turn this value into a global latitude & longitude (or global indices). @klindsay28 pointed out an issue in the POP driver that we were not checking for errors after `set_interior_forcing()` returns; ideally the solution to this issue would be an update to `marbl_logging_type` in a fashion that could be used by both the surface and interior forcing routines (especially if we settle #81 by deciding to pass multiple elements to `set_interior_forcing`)

mnlevy1981:
On the MARBL side, we should set `ElementID` to -1 by default, and then POP can automatically append lat / lon information given a positive ID. (Perhaps `element_ind` would be a better variable name.)

Also on the MARBL side, `lall_tasks` is a confusing variable name. Instead we should flip the meaning of it and use `lonly_master_writes`. By "flip the meaning" I mean that
```
lonly_master_writes = .not. lall_tasks
```
So we need to add (or remove) `.not.` from all uses.

mnlevy1981:
Note that we also need POP's `print_marbl_log()` function to be aware of when it's being called -- after `set_surface_forcing` the element index corresponds to a specific column in the block, but after `set_interior_forcing` it refers to a specific level of a column. In the latter case, we will need to provide `print_marbl_log` with either the lat / lon of the column or the `(i,j,bid)` triplet. `bid` is already required, but one possibility is to make `i` and `j` optional arguments that are only passed after the interior call.

mnlevy1981:
I have a MARBL branch that takes care of the issues in my second comment ('lall_tasks' => 'lonly_master_writes' and setting 'ElementInd = -1'); working on the POP modifications following a conversation with @klindsay28 

We are going to store `marbl_col_cnt` as well as mappings between `index_marbl` and `(i,j)` so we can map between surface elements and POP columns; for the latter, I realized that we can use `marbl_col_cnt` as `gcm_num_elements_surface_forcing` if we initialize things in the right order (i.e. determine `marbl_col_cnt` before initializing MARBL).

mnlevy1981:
I have a first pass at improved logging; changing the width of the bounding bracket to force a failure in `drtsafe` leads to messages like

```
[...]

PR #149, opened 2017-04-11
Bugfix/better logging

mnlevy1981:
A couple of improvements to message logging:

1. Clearer variable name (`lonly_master_writes = .true.` instead of `lall_tasks = .false.` if a message only needs to be written once)
1. Bugfix in logging location of error from `marbl_co2calc_mod::drtsafe` -- had previously been referencing `c` outside of the `do`-loop it indexed

Also tried to add a few comments to explain what was being added to the log in `drtsafe()`


mnlevy1981:
This pull request addresses issue #99

Issue #137, opened 2017-02-21
non_living_biomass_ecosys_tracer_cnt is not needed


mnlevy1981:
We have a local variable in `marbl_mod:marbl_init_tracer_metadata` that is an intent(out) from `marbl_mod:marbl_init_non_autotroph_tracer_metadata` but then never referenced. I propose removing the variable from `marbl_init_tracer_metadata()` and making it local to `marbl_init_non_autotroph_tracer_metadata()` rather than part of the interface.

mnlevy1981:
Talking with Keith, we realized that even keeping `non_living_biomass_ecosys_tracer_cnt` local, it imposes a restriction on the tracer indexing that we don't want... so it should be removed altogether.

mnlevy1981:
Fixed in marbl0.16.1

PR #165, opened 2017-05-26
Enhancement/remove non living tracer cnt

mnlevy1981:
Remove the `non_living_biomass_ecosys_tracer_cnt` variable (see issue #137).

Testing: verified that the tracer names and metadata did not change with this refactor by running a single POP CECO (+ciso) comparison test on both Cheyenne (intel 17) and Hobart (nag); this testing did catch some initial typos, and one of the errors flagged (I had accidentally changed the `ALK` tracer to `alk`) also caught a typo in a MARBL error traceback.

### DIFF
--- a/src/marbl_interface.F90
+++ b/src/marbl_interface.F90
@@ -559,7 +559,7 @@ contains
     call this%interior_forcing_ind%construct(this%tracer_metadata%short_name, &
                                    tracer_restore_vars, this%StatusLog)
     if (this%StatusLog%labort_marbl) then
-      call this%StatusLog%log_error_trace("itnerior_forcing_ind%construct",   &
+      call this%StatusLog%log_error_trace("interior_forcing_ind%construct",   &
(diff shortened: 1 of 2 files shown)


klindsay28:
If the comparison
`if (n == alk_ind .or. n == alk_alt_co2_ind) then`
in `subroutine marbl_init_non_autotroph_tracer_metadata` is changed to
`if (short_name == 'ALK' .or. short_name == 'ALK_ALT_CO2') then`
then the arguments `marbl_tracer_indices` and `n` can both be removed from the subroutine.

At that point, I call into question the need for (and usefulness of) the `associate` construct in `subroutine marbl_init_non_autotroph_tracers_metadata`. Each individual tracer index is only used once, and as you get deeper into the sequence of calls to `marbl_init_non_autotroph_tracer_metadata`,  you separate the usage of the tracer index from its association into `marbl_tracer_indices`.

PR #169, opened 2017-06-01
Introduce local-pgi machine

mnlevy1981:
PGI offers Community Edition of PGI 17.4, which can be installed locally and
then used to build / run the stand-alone MARBL tests. The default machine is
still 'local-gnu'; it might be time to consider adding a `~/.marbl/` directory
and letting users specify what compilers are available if the machine is not
recognized (so default machine would be 'local', and if `~/.marbl/` is not found
then only gnu would be supported; otherwise users could specify other compilers
that are installed)

** Compiler available at http://www.pgroup.com/products/community.htm

Verified that this does not change behavior on cheyenne or hobart

### DIFF
--- a/tests/python_for_tests/machines.py
+++ b/tests/python_for_tests/machines.py
@@ -3,6 +3,7 @@ from os import system as sh_command
 
 # Supported machines for running MARBL tests
 supported_machines = ['local-gnu', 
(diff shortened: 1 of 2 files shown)


klindsay28 (approved):
looks good

Issue #66, opened 2016-05-19
Abort if gcm_num_elements_interior_forcing is not 1?


mnlevy1981:
We have hard-coded assumptions about `num_elements` in `marbl_mod:marbl_set_interior_forcing` (and possibly elsewhere) such that we can not run with multiple interior elements. Until we fix these issues, I think we want to abort if the driver tells `marbl_instance%init` that it is trying to run with more than one interior element.

mnlevy1981:
Relatedly, we have added a `num_elements` dimension to some variables related to interior forcing but not to others and should maybe consider extending that to _all_ variables that would need it... on the other hand, we've also discussed removing `num_elements` from the interior forcing altogether since operating on multiple columns simultaneously will not necessarily improve performance when columns have different numbers of active levels.

mnlevy1981:
One idea that came up in discussion on May 30th was to remove `gcm_num_interior_forcing_elements` from the `init()` interface and just hard-code it to 1 until we are ready to allow it to be variable (or we decide to remove it altogether*).

That would not fully address this issue, but would be enough of a work-around that we could postpone further improvements until after MARBL 1.0.0 is released.

* Given that ROMS (and possibly some of the vertical coordinates available in MOM) use `KMT = km` for every cell, it is likely that we will want to consider letting MARBL handle multiple columns simultaneously.

mnlevy1981:
As code stands now (for MARBL 1.0.0), we force `num_interior_forcing_elements = 1` and don't offer a way to change it via the interface; after MARBL 1.0.0 is released we should revisit this and update the code to allow the GCM to pass multiple columns to MARBL.

PR #170, opened 2017-06-02
gcm_num_elements_interior_forcing not in interface

mnlevy1981:
The implementation of gcm_num_elements_interior_forcing is not complete, and if
this variable is not set to 1 then MARBL will produce unexpected results or
errors. For now, I have removed this variable from the interface to ensure that
it is always set to 1 inside the library.


mnlevy1981:
This should let us close #66 and then after MARBL 1.0.0 we could look into what needs to be done to allow multiple interior columns

PR #171, opened 2017-06-04
Had some bad whitespace in log

mnlevy1981:
3 unnecessary write statements were adding an extra empty line to the log, and
there was also one spot where I should have used the log_header() routine but
didn't.

(I was playing with MARBL on a machine I don't often use, and these issues stood out)


klindsay28 (approved):
looks good

Issue #86, opened 2016-07-15
Split up marbl_mod.F90


mnlevy1981:
`marbl_mod.F90` could likely be refactored into
1. `marbl_surface_forcing_mod.F90`
2. `marbl_interior_forcing_mod.F90`
3. `marbl_setup_mod.F90` (for configuration and initialization routines)

This would be an aesthetic fix, making the code significantly more readable / navigable. It would likely not have an effect on the interface, so we can decide if we want to go down this path post MARBL 1.0.0.

mnlevy1981:
Don't want `_forcing` in the module name

mnlevy1981:
I have a branch where I split `marbl_mod.F90` into `marbl_surface_mod.F90`, `marbl_interior_mod.F90`, and `marbl_glo_avg_mod.F90`. I still need to come up with better names for `set_surface_forcing()` and `set_interior_forcing()` -- I was thinking `marbl_surface_fluxes_compute()` and `marbl_interior_tendencies_compute()`? `marbl_surface_compute_fluxes()` and `marbl_interior_compute_tendencies()` is another option, though I don't really like splitting the phrases "surface fluxes" and "interior tendencies" like that.

Note that `marbl_interior_mod.F90` also has `marbl_set_global_scalars_interior()`, which I could see naming `marbl_interior_set_global_scalars()` or I can move it into `marbl_glo_avg_mod` (`marbl_glo_avg_set_interior_scalars()`?). I haven't done that because the subroutine basically just calls `adjust_bury_coeff()`, which I think belongs in `marbl_interior_mod.F90`.

mnlevy1981:
Now I'm leaning towards `marbl_interior_compute_tendencies()`. I think that works best with the other subroutines in interacts with -- on the interface, we'll have `marbl_interface%compute_tendencies()`, and for the carbon isotope tracers we'll have `marbl_ciso_compute_tendencies()`. Similarly, we can have `marbl_surface_compute_fluxes()`, `marbl_interface%compute_fluxes()`, and `marbl_ciso_compute_fluxes()`.

Note that splitting `marbl_mod` introduces a discrepancy between how we treat the base tracers (split between `marbl_surface_mod` and `marbl_interior_mod`) and how we treat the carbon isotopes. Should `marbl_ciso_mod` be split into `marbl_ciso_init_mod`, `marbl_ciso_surface_mod`, and `marbl_ciso_interior_mod`?

mnlevy1981:
[...]

PR #177, opened 2017-06-07
ciso_interior_forcing shouldn't set DIC_ & DOC_loc

mnlevy1981:
The DIC_loc_fields and DOC_loc_fields components of
marbl_interior_share_type are initialized to c0 below level kmt in
marbl_setup_local_tracers() [since all values in tracer_local = c0 for
k>kmt]. By not setting these two components to c0 in marbl_ciso_mod, we
can change the intent of the interior_share argument from intent(inout)
to intent(in).

### DIFF
--- a/src/marbl_ciso_mod.F90
+++ b/src/marbl_ciso_mod.F90
@@ -238,7 +238,7 @@ contains
     type(marbl_interior_share_type)         , intent(in)    :: marbl_interior_share(marbl_domain%km)
     type(marbl_zooplankton_share_type)      , intent(in)    :: marbl_zooplankton_share(zooplankton_cnt, marbl_domain%km)
     type(marbl_autotroph_share_type)        , intent(in)    :: marbl_autotroph_share(autotroph_cnt, marbl_domain%km)
-    type(marbl_particulate_share_type)      , intent(inout) :: marbl_particulate_share
+    type(marbl_particulate_share_type)      , intent(in)    :: marbl_particulate_share
     real (r8)                               , intent(in)    :: temperature(:)
     real (r8)                               , intent(in)    :: column_tracer(:,:)
     real (r8)                               , intent(inout) :: column_dtracer(:,:)  ! computed source/sink terms (inout because we don't touch non-ciso tracers)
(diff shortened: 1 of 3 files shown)


mnlevy1981:
Note the change to `marbl_mod.F90` (stripping a comment): we actually need `marbl_interior_share` to be `intent(inout)` in `marbl_set_interior_forcing()` because it gets passed to `marbl_export_interior_shared_variables()` as an `intent(inout)` variable in line 1216; the comment was incorrect.

klindsay28:
Could `marbl_interior_share`, and the other _share vars for that matter, be local variables in `marbl_set_interior_forcing` of dimension `domain%km`? I'm mentioning this because I don't see these vars being used outside of `marbl_set_interior_forcing` and its callees. I guess we might not want to do that if

1. there is a performance hit from the allocation going on (happening automatically under the covers),
2. we envision needing it outside of `marbl_set_interior_forcing`

mnlevy1981:
I think that would be a good change to make -- in my mind, components of the `marbl_interface_class` either need to be available to the GCM or are part of the interface for thread safety (that is probably not a complete list, but I doubt any categories I may be missing apply to `interior_share`). Moving `interior_share` to a local variable would keep thread safety, and the GCM certainly doesn't need to access this derived type.

I had originally thought that we should wait until after we resolve #86 because our split of `marbl_mod.F90` (and, I imagine, `marbl_ciso_mod.F90`) could potential result in calling `marbl_ciso_set_interior_forcing()` from the `marbl_interface:set_interior_forcing` instead of from `marbl_set_interior_forcing()`... but then we would just declare the local variable in the interface routine instead... so I'll implement this suggestion and push the update if there is not a performance hit.

mnlevy1981:
It looks like `particulate_share` and `surface_forcing_share` need to stay in the interface because of how they are initialized, but `interior_share`, `zooplankton_share`, and `autotroph_share` do not benefit from being part of the interface so they have been moved to local variables in `marbl_set_interior_forcing`.

This may change if we move the `domain%km` dimension inside the type -- we have talked about doing that for some datatypes but I don't remember off the top of my head if these share types are included in that design. For now, though, I think taking these variables out of the interface is a good decision.

klindsay28 (approved):
looks good

PR #178, opened 2017-06-08
marbl_status_log no longer optional in drtsafe

mnlevy1981:
There was a period of time where POP's abio tracer module was using MARBL's
solver, which meant we needed to be able to call drtsafe without a
marbl_status_log. Since that is no longer the case, and since omitting the
status log would mean that MARBL would error out silently, it makes more sense
to require this argument to the subroutine.

Note that all calls to drtsafe from inside MARBL already pass this optional
argument, so these changes were contained to drtsafe() itself.


klindsay28 (approved):
looks good

PR #180, opened 2017-06-08
CaCO3_form instead of CaCO3_prod

mnlevy1981:
This name change applies to the components of `marbl_autotroph_share_type` and
`autotroph_secondary_species_type`, and was done to ensure consistency between
internal variable names and the variable names provided in diagnostic output.


klindsay28 (approved):
looks good

PR #182, opened 2017-06-09
marbl_compute_PAR shouldn't compute totalChl

mnlevy1981:
Introduce `totalChl_local` to `marbl_set_interior_forcing()`, compute it in
`marbl_setup_local_tracers()`, and then pass it to `marbl_compute_PAR()`

Note that `marbl_consistency_check_autotrophs()` is now called from
`marbl_setup_local_tracers()` to ensure `totalChl_local` is correct.

### DIFF
--- a/src/marbl_mod.F90
+++ b/src/marbl_mod.F90
@@ -1092,9 +1092,6 @@ contains
        return
     end if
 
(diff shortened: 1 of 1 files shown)


klindsay28 (approved):
looks good

PR #183, opened 2017-06-12
Consistent use of molw_Fe constant

mnlevy1981:
marbl_constants_mod has a parameter for the molecular weight of iron, but
marbl_parms was not using the parameter in a computation involving the
constant. Also, many comments regarding the constant were missing information
or contained un-necessary details.


klindsay28 (approved):
looks good

PR #185, opened 2017-06-13
Enhancement/nl buffer update

mnlevy1981:
MARBL doesn't need to provide the GCM with `marbl_nl_in_size`, `marbl_nl_cnt`, or `marbl_nl_buffer_size` -- the GCM should determine these sizes and then MARBL can figure them out from `size()` and `len()`of various `intent(in)` strings.

Since the GCM no longer gets these values from MARBL, there is [an associated POP tag](https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n97_marbl_dev_n48_cesm_pop_2_1_20170601).

### DIFF
--- a/src/marbl_ciso_mod.F90
+++ b/src/marbl_ciso_mod.F90
@@ -102,8 +102,8 @@ contains
     !-----------------------------------------------------------------------
     !  local variables
     !-----------------------------------------------------------------------
-    character(*), parameter :: subname = 'marbl_ciso_mod:marbl_ciso_init_tracer_metadata'
-    character(len=char_len) :: log_message
+    character(len=*), parameter :: subname = 'marbl_ciso_mod:marbl_ciso_init_tracer_metadata'
+    character(len=char_len)     :: log_message
 
     integer (int_kind) :: n                             ! tracer index
     integer (int_kind) :: auto_ind                      ! autotroph functional group index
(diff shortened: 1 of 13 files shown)


mnlevy1981:
After the code review this morning, I implemented the two changes we talked about:

1. no continuation line for declaring `array_out`
1. using `character(len=___)` format instead of `character(___)`
   * also, subroutine arguments with explicit `intent()` should all be `character(len=*)`

I'm a little worried that I carried away in the second task, and should have just stuck to the `intent` variables, because now this is a pretty intrusive commit.

There were also updates in POP, resulting in a [new tag](https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n99_marbl_dev_n48_cesm_pop_2_1_20170601):

```
$ svn diff \
https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n96_marbl_dev_n48_cesm_pop_2_1_20170601 \
https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n99_marbl_dev_n48_cesm_pop_2_1_20170601
```

klindsay28 (approved):
looks good

Issue #184, opened 2017-06-13
Make autotroph, zooplankton, and grazing counts runtime settable


mnlevy1981:
This is a big step towards removing all build-time settings from MARBL. The best path forward looks to be

1. Clean up `marbl_sizes.F90` so that `tracer_restore_cnt`, `num_surface_forcing_fields`, and `num_interior_forcing_fields` are local variables in the `*_forcing_indexing_type` constructors
   * outside of the constructors, use of these variables can be replaced with `size(___)`
   * We can allocate `this%tracer_id` to `tracer_restore_cnt` instead of `marbl_total_tracer_cnt`

1. introduce a fourth phase initialization; the phases should be named

   ```
   marbl_init_sizes() ! NEW ROUTINE
   marbl_init_configuration() ! formerly marbl_config
   marbl_init_parameters_and_tracers() ! formerly marbl_init
   marbl_init_complete() ! formerly marbl_complete_config_and_init
   ```

   * First phase will set `autotroph_cnt`, `zooplankton_cnt`, and `grazer_prey_cnt`; completing that phase lets us allocate `autotrophs_config`, `autotrophs`, etc etc

Next step towards runtime configurability will removing the `parameter` attribute from `ecosys_base_tracer_cnt` and `ciso_tracer_cnt`

mnlevy1981:
I think `num_surface_forcing_fields` and `num_interior_forcing_fields` will actually need to be local to `complete_config_and_init()` and `intent(out)` from the constructors, I typed this up a little too hastily.

mnlevy1981:
After talking to @matt-long another approach to initialization would be to have some way to buffer `put` calls and then all GCMs (whether using namelist or not) could call the `init_all_phases()` routine and the individual phases could be private routines. I think most of the infrastructure to do this already exists and we can rely on a new instance of `marbl_config_and_parms_type`, with some tweaks to error conditions in the `put` call itself. We'll discuss this more at the [July 25 meeting](https://github.com/NCAR/MARBL/wiki/2017-MARBL-Dev-team-meetings#july-25-2017).

mnlevy1981:
`autotroph_cnt`, `zooplankton_cnt`, and `max_grazer_prey_cnt` can now be set at runtime via inputfile; you must also set

`PFT_default = 'user-specified'`

and then provide complete parameterization for all PFTs.

Issue #46, opened 2016-04-18
We need a build-namelist tool


mnlevy1981:
How do the GCMs produce namelists to be sent to MARBL? Does it make sense to have a build-namelist utility in MARBL? One option would be to produce XML files for definitions / default values as is done in CESM / ACME components.

mnlevy1981:
yes, we need this tool

mnlevy1981:
Note: it doesn't make sense to start work on this until we have closed #56 and #75 because those two issues each involve moving a MARBL namelist to the driver.

---

Talked with @quantheory and @mvertens -- Sean is close to finishing the python-ization of the `build-namelist` tool for the data models in CIME, and I think we can leverage a lot of that work. It sounds like the starting point will be:
1. use Sean's `namelist.py`, `namelist_defaults.py`, and `namelist_definitions.py` for defining namelist objects and parsing the defaults / definition files (perhaps removing CIME-specific functionality that we don't need).
   1. There are some slight dependencies on CIME scripts in those files, but the only ones that we would want to MARBL-ize are the error-handling calls
   2. We may want to update `namelist_defaults.py` to handle the XML schema [Mariana prototyped](https://svn-ccsm-models.cgd.ucar.edu/pop2/branches/newbldnamelist/bld/namelist_files/namelist_defaults_pop2.xml)
2. Write our own `build-namelist` tool using the datm version of the file as a template (but there's a lot of code in that script that MARBL does not need, i.e. reading in stream files)

We also brain-stormed ways to handle the namelist object in Python. I think we want something similar to `marbl_single_config_or_parm_type` in [marbl_config_mod.F90](https://github.com/NCAR/MARBL/blob/marbl0.8.0/src/marbl_config_mod.F90#L173), where we only define `string`, `logical`, `real`, and `integer` variables (so we turn arrays and derived types into a several individual variables). Perhaps the biggest obstacle we will face from this is parsing something like

`grazing%sname = 'foo'`

in the `user_nl_marbl` file, because `grazing` is an array so we will need to convert that to

`grazing(1,1)%sname = 'foo'`

(or whatever Fortran would treat it as were it reading a namelist).

mnlevy1981:
@klindsay28 pointed out that the `build-namelist` tool should be able to output in some non-namelist formats as well (something the MPAS registry can parse and something the MOM6 parms file can parse both come to mind, other GCMs may have other parameter input mechanisms).

mnlevy1981:
[...]

PR #189, opened 2017-07-13
Enhancement/runtime pft count

mnlevy1981:
On my way to runtime PFT counts, I realized it wasn't that difficult to get runtime tracer counts... but I'd like to talk about how I achieved that before pressing forward with the PFT count variables. With the changes from today, finishing #184 will let us remove `marbl_sizes.F90`.

This is already a pretty crowded pull request:
* the first couple commits (5ec676b and fde5862) address the first point of #184 (clean up `marbl_sizes.F90`)
* The next commit (1a0b842) addresses the first half of the second point (renames existing setup routines but doesn't introduce the new routine)
* The following commit (6afefd0) introduces an new `init()` routine on the interface that calls the three phases in order, using a namelist to initialize config vars / parameters. For GCMs that want to use the namelist instead of `%put` and `%get`, this is easier than calling the three phases individually
* The next two commits (b6c9a22 and 14efae6) diverge from the goals of #184 and make the tracer count entirely runtime configurable (no more `ecosys_base_tracer_cnt` or `ciso_tracer_cnt` parameters in `marbl_sizes`)

There are some small POP changes as well. Development started from [marbl_dev_levy_n105](https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n105_marbl_dev_n54_cesm_pop_2_1_20170623/); [marbl_dev_levy_n106](https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n106_marbl_dev_n54_cesm_pop_2_1_20170623/) uses the single init routine from 6afefd0 and [marbl_dev_levy_n107](https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n107_marbl_dev_n54_cesm_pop_2_1_20170623/) relies on POP's `buildcpp` script instead of MARBL for tracer counts (so `ECOSYS_BASE_NT` is still a build-time macro, and I also introduced `CISO_NT=14`, but they are used by POP rather than MARBL)


mnlevy1981:
@klindsay28 -- the eight commits between (and including) 9ee2746 through 66c7bfb deal with modifications that came up in our code review. There are two things I did NOT address, because I think next week's meeting will clarify the best path forward:

1. I did not rename the catch-all initialization with the required `gcm_nl_buffer` argument to `init_allphases()`; it is still `init()` for now
1. I did not introduce an easy way for GCMs to call `init()` with the default namelist

It turns out you can not use

```
   integer, dimension(size(A)) :: B
```

or

```
   integer, dimension(shape(A)) :: B
```

in place of

```
   integer, dimension(size(A,1), size(A,2)) :: B
```

Which we had talked about briefly, I believe for making sure `tracers_local` and `tracers` had the same dimensions.

(No action necessary right now, just leaving a note so we know where to pick up at the next review... I'm going to get back to variable PFT count now and wanted to make sure it was clear where that work began)

mnlevy1981:
Based on [Tuesday's meeting](https://github.com/NCAR/MARBL/wiki/2017-meeting-notes#july-25-2017), I'm planning on the following path forward:

1. All parameter setup should be in `marbl_parms.F90` => merge `marbl_config_mod.F90` and `marbl_parms.F90`
1. Rework `config_and_parms_type` so that `marbl_instance%user_settings%put()` can be called before any of the `define_variable()` routines

   * This means merging `%configuration` and `%parameters` into `%user_settings`, so the locking mechanism and error criteria of this type will change

1. One monolithic `marbl_instance%init()` instead of arbitrary `init_phase#()`

   * Once all the `put()` statements occur before we define the variables, we no longer need to pause between phases to give the GCM the opportunity to call `put()`

1. POP's `build-namelist` will provide a new (text) file that can be slurped up and turned into `user_settings%put()` calls

   * Format to be determined later, but probably one variable per line, with each line containing
   ```
   varname  datatype  value
   ```
   * This will also be a good first step towards dealing with #46

1. remove all namelists from MARBL
1. Add third category of parameters (`configuration_vars`, `parameters`, and `PFT_parameters`? Names will be entirely internal to MARBL, but should be easy for future developers to determine where new variables belong)

   * move `*_config` (`autotrophs_config`, etc) from `config_vars` to `parameters` (and maybe rename these variables `*_define`?)
   * move PFT-specific variables from `parameters` to `PFT_parameters`

1. Move PFT count variables from being build-time specified in `marbl_sizes.F90` to run-time specified in `configuration_vars`

   * Will let us remove `marbl_sizes.F90`
   * Will require allocated memory for `*_config` (similar to how we allocate memory for `tracer_restore_vars` after constructing the tracer indices)

I'm open to suggestions on a better order to do things, but am getting started on merging `marbl_config_mod` and `marbl_parms` now.

mnlevy1981:
To avoid some strings being read as real-valued variables (and value = `NaN`), 9485bdf uses the `ieee_is_nan()` function out of `ieee_arithmetic` (`pgf90` doesn't recognize `isnan()`); this is not available in gfortran v4, so we now have two minimum compiler requirements that need to be documented somewhere:

gfortran: v5 or later
PGI: v17.04 or later

mnlevy1981:
As of 27d8837 and [marbl_dev_levy_n112](https://svn-ccsm-models.cgd.ucar.edu/pop2/branch_tags/marbl_dev_levy_tags/marbl_dev_levy_n112_marbl_dev_n54_cesm_pop_2_1_20170623), I have made it through item 5 of the list posted on [July 27](https://github.com/NCAR/MARBL/pull/189#issuecomment-318416961) -- MARBL has a `parse_inputfile_line()` routine on its interface that takes a string like

```
varname = value
```

and returns strings containing `varname`, `value`, and a guess at `datatype`. For example, if the line reads

```
parm_fe_bioavail = 1.1
```

then the subroutine returns `varname = 'parm_fe_bioavail'`, `value = '1.1'`, and `datatype = 'real'`. There is also a new procedure as part of the `marbl_instance%put_setting()` interface that takes a `varname`, `datatype`, and `value` as arguments, converts `value` from a string type to whatever is specified in `datatype`, and then calls `put_setting()` again.

Besides ignoring comments (anything including / following a `!`), `parse_inputfile_line()` also ignores lines that begin with`&` or lines that only contain `/`; this means a standard Fortran namelist can be used as a MARBL input file, so POP's `build-namelist` tool was updated to create `marbl_in` as well as `pop_in` but `marbl_in` does not need any special formatting treatment.

One issue I've come across is how to determine if a value is an integer or a real number. I initially used

```
read(value, *, iostat=ioerr) ival
```

and relied on `ioerr .ne. 0` to alert me to non-integer values. However, the line

```
read('3.5', *, iostat=ioerr) ival
```

returns `ioerr = 0` with the Intel and PGI compilers (`ival = 3`) while NAG and GNU both return positive `ioerr`. So instead the check for integer value is tied to the check for real values:

```
read(value, *, iostat=ioerr) rval
```

and if `ioerr .eq. 0` then `rval .eq. floor(rval)` => integer, otherwise it is real. I'm not sure that's the best approach (you can set an integer variable with `value = 1.0` in the inputfile), so we should discuss other options for parsing lines at the Aug 15 meeting.

And regardless of how we determine datatype, I still need to refactor the checks because there is a lot of code duplicated between `parse_inputfile_line()` and `put_all_string()`, not to mention the array-handling is pretty clunky. But I'd like to actually get runtime-configurable PFT counts up and running (and the associated driver changes) before coming back to clean all this up.

Issue #196, opened 2017-09-20
Change how namelist / inputfile are passed to stand-alone driver


mnlevy1981:
Instead of hard-coding `test.nml` as the namelist name and using ` < ` to pipe the inputfile via stdin, we should use command-line arguments (`-i [inputfile] -n [namelist]` or similar).

This is necessary based on a discussion about [using the stand-alone driver in CESM case generation](../wiki/2017-meeting-notes#september-19-2017):

1. We will need to use at least two different namelists (one to compute tracer count before POP is built, and one to generate the input file)
1. We may want to add flexibility by allowing users to pass multiple input files - for example, call `put_setting()` with a resolution-dependent input file and then with the user-provided input file.

mnlevy1981:
fixed in [marbl0.20.3](https://github.com/NCAR/MARBL/releases/tag/marbl0.20.3)

PR #197, opened 2017-09-21
Enhancement/commandline args

mnlevy1981:
Rather than hard-coding the namelist as test.nml and reading the inputfile from
stdin, the marbl driver now expects the namelist file name to be the first
command line argument and if a second argument is present then that is the
inputfile (otherwise lhas_inputfile = .false. and nothing is read)

Also, only task 0 reads / writes files when running with multiple mpi tasks.


mnlevy1981:
Addresses #196

PR #198, opened 2017-09-21
Add get_tracer_cnt routine to interface and tests

mnlevy1981:
The get_tracer_cnt test returns the number of tracers in the ecosys_base
module, the number of tracers in the ciso module, and the total MARBL tracer
cnt all in one line, separated by spaces. So with ciso_on = .true. and all
other parameters kept to default settings, the output of the test is

Beginning tracer_cnt test...
32 14 46

One use for this test will be for POP to determine ECOSYS_BASE_NT, CISO_NT, and
MARBL_NT directly from the MARBL code rather than relying on users to update it
correctly.

### DIFF
--- a/src/marbl_interface.F90
+++ b/src/marbl_interface.F90
@@ -121,6 +121,7 @@ module marbl_interface
      procedure, public  :: extract_timing
      procedure, private :: glo_vars_init
      procedure, public  :: get_tracer_index
+     procedure, public  :: get_tracer_cnt
      procedure, public  :: set_interior_forcing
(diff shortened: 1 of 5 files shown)


mnlevy1981:
This test isn't necessary at the moment, and there were some questions about the functionality of `get_tracer_cnt` (should it just be `get_module_tracer_cnt` since there are other ways of getting total tracer count?); rather than spending time / energy determining the best implementation, @klindsay28 and I decided it would be better to just not merge it.

Issue #208, opened 2017-12-04
MARBL_tools should use case-insensitive string compares


mnlevy1981:
When generating a `MARBL_settings_class` object, case should not matter for the variable names in the user-provided setting input file.

We have a Fortran-based string comparison in `src/marbl_settings_mod.F90` (the function `case_insensitive_eq`) but we need something similar for the Python code as well.

mnlevy1981:
Rather than case-insensitive comparisons, I just use `varname.lower()` for keys in the input file dictionary

mnlevy1981:
Fixed in [marbl0.26.2](https://github.com/NCAR/MARBL/releases/tag/marbl0.26.2)

PR #209, opened 2017-12-04
Two python improvements

mnlevy1981:
1. MARBL_tools/__init__.py is now compatible with python3
2. The keys for the dictionary representing variables read from user-provided
   input file are case-insensitive, so variables don't need to match the exact
   format used in default_settings.json; this is useful because Fortran is case
   insensitive and we want to mimic the behavior of Fortran namelists.

This addresses #208; note that, even after updating `__init__.py`, `MARBL_tools` requires python 2. When running `MARBL_generate_settings_file.py` with python 3, I get an error because the python 2 types `str` and `unicode` have been renamed `bytes` and `str`, respectively.


mnlevy1981:
Second commit fixes issue with python 3 (was a quick fix that came to me out of the blue; figured I'd update this branch instead of risk forgetting about it)

mnlevy1981:
All python scripts in MARBL can run with either python v2 (2.6 or later, I believe... maybe 2.7?) or v3. I think we want to get this on to master before CESM 2.

mnlevy1981:
Running `aux_pop_MARBL` on both hobart and cheyenne to make sure changes following the code review don't break the CESM workflow (I'm fairly confident they won't); assuming the tests all pass I'll accept this pull request and make a new `marbl_dev` tag of POP tonight or over the weekend.

Issue #226, opened 2018-01-29
Un-necessary CISO diagnostics


mnlevy1981:
The following are defined for all autotrophs, but are only needed for calcifiers:

```
lname    = trim(autotrophs(n)%lname) // ' Ca13CO3 Formation'
sname    = 'CISO_' // trim(autotrophs(n)%sname) // '_Ca13CO3_form'

lname    = trim(autotrophs(n)%lname) // ' Ca13CO3 Formation Vertical Integral'
sname    = 'CISO_' // trim(autotrophs(n)%sname) // '_Ca13CO3_form_zint'

lname    = trim(autotrophs(n)%lname) // ' Ca14CO3 Formation'
sname    = 'CISO_' // trim(autotrophs(n)%sname) // '_Ca14CO3_form'

lname    = trim(autotrophs(n)%lname) // ' Ca14CO3 Formation Vertical Integral'
sname    = 'CISO_' // trim(autotrophs(n)%sname) // '_Ca14CO3_form_zint'

lname    = trim(autotrophs(n)%lname) // ' d13C of CaCO3'
sname    = 'CISO_autotrophCaCO3_d13C_' // trim(autotrophs(n)%sname)

lname    = trim(autotrophs(n)%lname) // ' d14C of CaCO3'
sname    = 'CISO_autotrophCaCO3_d14C_' // trim(autotrophs(n)%sname)
```

mnlevy1981:
Fixed in [marbl0.26.0](https://github.com/NCAR/MARBL/releases/tag/marbl0.26.0)

Issue #200, opened 2017-09-22
Do we need tracer count as an optional intent(out) from init?


mnlevy1981:
After calling `marbl_instance%init()`, GCMs can just query `size(marbl_instance%tracer_metadata)` to find the tracer count.

mnlevy1981:
Note that POP no longer uses the optional intent(out) from `init()` so this can be fixed without a new POP tag.

mnlevy1981:
Fixed in [marbl0.26.0](https://github.com/NCAR/MARBL/releases/tag/marbl0.26.0)

Issue #192, opened 2017-08-17
Specifying a single PAR sub-column should be sufficient


mnlevy1981:
In an email, @maltrud wrote:

> i assumed that if i specified 1 sub-column for PAR that i wouldn’t have to do anything special.  i don’t know why i assumed this (in retrospect), but it took me quite a while to track down my seg faults to undefined `PAR_col_fac`.  maybe it makes sense to not require the user to go through the steps of setting this equal to 1 with, instead, a check within marbl.

It should be straight-forward to define `PAR_col_frac = 1` if there is just a single sub-column

mnlevy1981:
Fixed in [marbl0.26.0](https://github.com/NCAR/MARBL/releases/tag/marbl0.26.0)

Issue #150, opened 2017-04-14
reconsider declaration of shallow_depth in compute_vertical_integrals in marbl_diagnostics_mod.F90


klindsay28:
current declaration is:
```
    real(kind=r8)      :: shallow_depth = 100.0e2_r8
```
this implies a save attribute, which we probably want to avoid in general, for thread safety.
One option is to add the parameter attribute, to indicate that this is never changed.

mnlevy1981:
I added `shallow_depth` as an optional `intent(in)` argument to `compute_vertical_integrals()`, and if the argument is not present then `100.0e2_r8` is used. This seemed like a good extension of the interface, but if it's deemed overkill then we can use the `parameter` attribute instead.

matt-long:
This seems like a good extension of the interface to me.

mnlevy1981:
Fixed in [marbl0.26.0](https://github.com/NCAR/MARBL/releases/tag/marbl0.26.0)

PR #229, opened 2018-02-13
Enhancement/low hanging fruit

### LINKED ISSUE #202: unexpected behavior when marbl_in has more than marbl_in_line_cnt lines
`namelist_split_by_line` doesn't do any error checking to ensure that `marbl_nl_buffer` is a large enough array to contain the full contents of `marbl_in`; instead it silently truncates the file after `marbl_in_line_cnt` lines.

At the very least, we need an error check to abort if the full file is not copied into the array. A better option would be for `marbl_nl_buffer` to be allocatable, and `namelist_split_by_line` can allocate memory for it after determining how big it is.

mnlevy1981:
This branch will contain fixes for #226, #200, #150, #192, and #82. The corresponding POP tag will fix #202.

Fixes for #226, #200, and #150 are already on this branch; I will comment when the other two issues are fixed as well.

### DIFF
--- a/src/marbl_ciso_mod.F90
+++ b/src/marbl_ciso_mod.F90
@@ -27,6 +27,9 @@ module marbl_ciso_mod
   use marbl_settings_mod, only : autotrophs
   use marbl_settings_mod, only : zooplankton
   use marbl_settings_mod, only : grazing
+  use marbl_settings_mod, only : caco3_bury_thres_iopt
+  use marbl_settings_mod, only : caco3_bury_thres_iopt_fixed_depth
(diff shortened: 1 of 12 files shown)


mnlevy1981:
Fix for #82 changes answers when `ciso_on = .true.`

mnlevy1981:
I don't have a good way to test the fix for #192 in 55d220d, POP uses at least two PAR subcolumns (simplest case is one subcolumn under atmosphere and the other under ice).

mnlevy1981:
Also added a fix for #228

Issue #243, opened 2018-03-09
pass tracer_local to marbl_ciso_set_interior_forcing


klindsay28:
simplifications that can then be done are
don't need `DIC_loc`, `O2_loc`, `NO3_loc` `DOCtot_loc` in `marbl_interior_share`
(compute `DOCtot_loc` in `marbl_ciso_set_interior_forcing`)
remove `setup_local_column_tracers`

investigate potential simplifications in `marbl_export_zooplankton_shared_variables`, `marbl_export_autotroph_shared_variables`, `setup_local_autotrophs`, `marbl_autotroph_consistency_check`

mnlevy1981:
fixed in marbl0.28.8

PR #263, opened 2018-04-11
Bugfix/240

### LINKED ISSUE #240: ciso dtracer computations don't include terms from grazing of zooplankton

mnlevy1981:
Added additional terms to `zooplankton_share_type` so that `marbl_ciso_mod.F90` could include `zoo_graze` terms when computing tracer tendencies. This fixes #240 in the "quick fix" sense, but @klindsay28 and I talked about how it makes sense for me to take a little extra time to clean up the way some of this data is passed from `marbl_mod.F90` to `marbl_ciso_mod.F90`. Namely:

1. Most (all?) of `autotroph_share_type` is a direct copy of `autotroph_secondary_species_type`; why don't we pass that instead of doing unnecessary memory copies?
1. `marbl_ciso_mod.F90` is only concerned with the sum over all zooplankton species, so this summation should be done as part of building `zooplankton_share_type` rather than done internal to `marbl_ciso_mod.F90`. That would be consistent with how `DOCtot` is handled (summed in `marbl_mod.F90`, treated as a single quantity throughout `marbl_ciso_mod.F90`)

I'll add another comment when this is actually ready for review.

### DIFF
--- a/autogenerated_src/default_diagnostics.json
+++ b/autogenerated_src/default_diagnostics.json
@@ -920,22 +920,22 @@
       "units": "mmol/m^3 cm/s",
       "vertical_grid": "none"
    },
-   "CISO_zooC_d13C": {
+   "CISO_zoototC_d13C": {
       "dependencies": {
          "ciso_on": ".true."
       },
       "frequency": "medium",
-      "longname": "d13C of zooC",
+      "longname": "d13C of total zooC",
       "operator": "average",
       "units": "permil",
       "vertical_grid": "layer_avg"
    },
-   "CISO_zooC_d14C": {
+   "CISO_zoototC_d14C": {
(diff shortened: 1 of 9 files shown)


mnlevy1981:
`marbl_zooplankton_share_type` is now used to pass the sum over all zooplankton of fields `marbl_ciso_mod.F90` needs. `marbl_autotroph_share_type` only passes tracer values, and `autotroph_secondary_species` is passed directly instead of being copied into other components of the share type. Note that the share type can be removed after we resolve #243.

mnlevy1981:
a284a46 addresses #243 (d700590 has most of the infrastructure, but a typo where `DO13Ctot_loc` was pointing to `DO14Ctot_loc`)

PR #267, opened 2018-04-25
add warning capability to marbl_logging, use it, and warning_trace, f…

klindsay28:
…or co2calc calls

Testing: marbl_dev_n85 and a4d31abedbaa1bccccb2f5f583bc9989dd08f701 on MARBL NCAR/master

aux_pop_MARBL hobart/nag: (baseline comparison to MARBL NCAR/master)
   except for tests mentioned below, all tests pass
      ciso tests failed RUN because marbl_dev wasn't up to date with MARBL NCAR/master

Files Modified:
	modified:   src/marbl_co2calc_mod.F90
	modified:   src/marbl_logging.F90
	modified:   src/marbl_mod.F90

### DIFF
--- a/src/marbl_co2calc_mod.F90
+++ b/src/marbl_co2calc_mod.F90
@@ -888,20 +888,17 @@ contains
 
              ! Iteration number
              WRITE(log_message,"(3A,1X,A,I0)") '(', subname, ')', 'it = ', it
-             call marbl_status_log%log_noerror(log_message, subname, c,       &
-                             lonly_master_writes=.false.)
+             call marbl_status_log%log_warning(log_message, subname, c)
 
              ! x1 & f
              WRITE(log_message,"(3A,1X,A,2E15.7e3)") '(', subname, ')',       &
                   'x1,f = ', x1(c), flo(c)
-             call marbl_status_log%log_noerror(log_message, subname, c,       &
-                             lonly_master_writes=.false.)
+             call marbl_status_log%log_warning(log_message, subname, c)
 
              ! x2 & f
              WRITE(log_message,"(3A,1X,A,2E15.7e3)") '(', subname, ')',       &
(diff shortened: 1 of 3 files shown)


mnlevy1981 (changes requested):
I'm a little confused about how the warning traceback works; if I remember correctly we are not mimicking the error traceback because the subroutine that triggers the warning does not return, so the log could potentially fill with misleading warning tracebacks. So do we want to set `lwarning = .false.` in `marbl_log_warning_trace()`? Otherwise a warning from `marbl_co2calc_surface() with flux_co2` might also print a warning pointing to `flux_alt_co2`, and a warning from `marbl_co2calc_interior() with dic` might also print a warning pointing to `dic_alt_co2`.

The other thing that stands out is that we now want to use `labort_on_warning` but there is no way for the user to turn that on. I think the best path forward to address that is for me to open up another issue ticket to address making `labort_on_warning` a new variable in `marbl_settings_mod.F90`.

So my suggestion is to add

```
     write(log_message, "(2A)") "Warning reported from ", trim(RoutineName)
     call this%log_warning(log_message, CodeLoc, ElemInd)
+    this%lwarning = .false.
```

to `marbl_log_warning_trace()` and then I'll address `labort_on_warning` separately. @klindsay28 how does that sound?

klindsay28:
I see your point about lwarning. A drawback to your suggestion is that the code can then only report a single level of warning trace. My initial usages of the warning interface only reports a single level, so your suggestion wouldn't cause problems for that, but it does limit what we can do in the future.

Alas, I can't think of an easy implementation that avoids spurious warnings and enables multiple warning trace entries.

In my opinion, avoiding the spurious warnings is preferable to the drawback. I'll put it in and update the pull request.

I agree that enabling labort_on_warning should be in a separate issue ticket.

PR #285, opened 2018-06-13
Change sphinx theme

mnlevy1981:
UCAR lawyers have pointed out that we can not have ads appearing on our documentation pages. If we want to continue to host documentation on [readthedocs.io](https://marbl.readthedocs.io/), then we can not use the `sphinx_rtd_theme` because it allows RTD to push ads. Instead, we switch to the `better` theme and also use `sphinxcontrib-fulltoc` to continue to get the full table of contents in the sidebar.


mnlevy1981:
Closing without merging - we went with the option presented in #284 instead

PR #298, opened 2018-07-13
New python tool to check formatting of fortran / python files

mnlevy1981:
As we start enforcing [coding standards](https://github.com/marbl-ecosys/MARBL/wiki/Code-Formatting-Rules-(STRAWMAN)), we need a way to check that developers are following the rules. This new script (which is run as part of the TravisCI testing) will house those tests.

I thought about using https://github.com/NCAR/code-format-checker or https://github.com/adcroft/house_cleaning but in the end it seemed easiest to just write something MARBL-specific: developers don't need to clone a different repository to access the tool, and we don't need to introduce command-line arguments to allow the flexibility of a more general tool.

I think this tool is in a good place to bring to `development` now, and then as I work through issues based on the coding standards I can just introduce new tests.


matt-long (approved):
@mnlevy1981 this looks good to me.

Issue #13, opened 2016-02-29
Consistent file names


mnlevy1981:
We want `module XYZ` in `XYZ.F90` (and `module ABC_mod` in `ABC_mod.F90`). I'm anticipating the following file name changes when we are ready:

```
$ mv ecosys_diagnostics_mod.F90 marbl_diagnostics_mod.F90 
$ mv ecosys_mod.F90 marbl_mod.F90
$ mv ecosys_restore.F90 ecosys_restore_mod.F90 
$ mv marbl_kinds.F90 marbl_kinds_mod.F90
$ mv marbl_share.F90 marbl_share_mod.F90
```

matt-long:
Should the ecosys_restore be marbl_restore?

---

Matthew C. Long
Scientist I, Oceanography Section
Climate and Global Dynamics Laboratory
National Center for Atmospheric Research
303.497.1311 (W)
650.575.7784 (M)
http://www.cgd.ucar.edu/staff/mclong

mnlevy1981:
It should - I was just making a list of what was needed to have the file name consistent with the module name for building, while ecosys_restore -> marbl_restore_mod would require updating the files that depend on ecosys_restore as well. That wasn't very clear in the issue ticket.

mnlevy1981:
The following are the only source files that are not `marbl_*_mod.F90` (they are missing the `_mod` suffix):

```
marbl_interface.F90
marbl_interface_constants.F90
marbl_interface_types.F90
marbl_internal_types.F90
marbl_logging.F90
marbl_oxygen.F90
marbl_parms.F90
marbl_sizes.F90
```

I don't know if we want to use `_mod` for ALL files, or if there is some criteria (maybe use `_mod` if the module has public subroutines or functions that are used by other modules?)

mnlevy1981:
It seems like there are three options to consider:

1. As mentioned [in 2016](https://github.com/marbl-ecosys/MARBL/issues/13#issuecomment-199401802), `marbl_XYZ.F90` => a module that only has public variables, whereas `marbl_XYZ_mod.F90` => a module with subroutines / functions
1. Every module should have the `_mod` suffix, e.g. `marbl_XYZ_mod.F90`
1. No module should have the `_mod` suffix, e.g. `marbl_XYZ.F90`

[...]

PR #303, opened 2018-07-27
File name clean-up

mnlevy1981:
We want all files to be `marbl_*_mod`, and they should contain `module marbl_*_mod`.

Only renamed six files, but all use statements from other files had to be changed accordingly. This will require changes in the GCM as well.

I also cleaned up the emacs-specific file header for formatting.


mnlevy1981:
I'm not going to merge this PR -- there should be some discussion on #13 about what we want the file names to be, and it makes sense to address #86 first rather than potentially needing to rename `marbl_mod.F90` before splitting it into two modules.

PR #306, opened 2018-07-27
Add -Wextra -Wno-compare-reals gfortran flags

mnlevy1981:
`-Wall` does not actually turn on ALL warnings, so we include `-Wextra` as well.
However, `-Wextra` flags checking for equality between two reals with the warning

```
Equality comparison for REAL(8) at (1)
```

And we want to allow those checks, so we also include `-Wno-compare-reals`.

These extra checks picked up on a few unused parameters (mostly `subname` being
declared in a subroutine and then never used)

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -108,7 +108,7 @@ all : gnu
 
 .PHONY: gnu
 gnu:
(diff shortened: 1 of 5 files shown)


klindsay28:
With the removal of these unused local variables, subroutines `marbl_diagnostics_set_surface_forcing` and `store_diagnostics_ciso_surface_forcing` now appear to have comments identifying local variables, but do not have local variables. I suggest removing these comment blocks.

Skimming through the code, I do not see any additional superfluous local variables comment blocks.

mnlevy1981:
@klindsay28 -- thanks for pointing out those unnecessary code blocks, I cleaned them up (as well as a few other misplaced comments). I think the changes here are still pretty small and it's not worth finding a time to sit together and go over them, but if you get a chance could you look over the changes from 1603831 and let me know if anything else stands out as missing? I've run tests on hobart and am currently also testing on cheyenne, though I'm pretty confident these won't change answers.

klindsay28 (approved):

PR #308, opened 2018-08-07
Split marbl_mod and marbl_ciso_mod

mnlevy1981:
Separate the initialization, surface flux computation, and interior tendency computations in both `marbl_mod.F90` and `marbl_ciso_mod.F90`: the end result is replacing those two modules with seven new ones and moving some routines to `marbl_init_mod.F90`:

1. `marbl_surface_flux_mod` -- contains `marbl_surface_flux_compute()`, formerly `marbl_set_surface_forcing()`
1. `marbl_interior_tendency_mod` -- contains `marbl_interior_tendency_compute()`, formerly `marbl_set_interior_forcing()`. Also has `marbl_set_global_scalars_interior()`, which needs to be renamed (ideally it would be moved to `marbl_glo_avg_mod` but it calls `adjust_bury_coeff()` which needs to stay in `marbl_interior_mod`)
1. `marbl_glo_avg_mod` -- contains `marbl_glo_avg_count_vars()` and `marbl_glo_avg_init_rmean_vals()`
1. `marbl_ciso_init_mod` -- contains `marbl_ciso_init_tracer_metadata()`
1. `marbl_ciso_surface_flux_mod` -- contains `marbl_ciso_surface_flux_compute()`, formerly `marbl_ciso_set_surface_forcing()`
1. `marbl_ciso_interior_tendency_mod` -- contains `marbl_ciso_interior_tendency_compute()`, formerly `marbl_ciso_set_interior_forcing()`.
1. `marbl_interior_share_mod` -- contains `marbl_interior_share_update_sinking_particle_from_level_above()`, which used to be a private routine (`marbl_update_sinking_particle_from_prior_level()`) in both `marbl_mod` and `marbl_ciso_mod`; now both tracer modules can use the same code. Perhaps this would also be a good place for `adjust_bury_coeff()`, allowing `marbl_set_global_scalars_interior()` to be moved to `marbl_glo_avg_mod`?

For consistency, the interface routines `set_surface_forcing()` and `set_interior_forcing()` have been renamed `surface_flux_compute()` and `interior_tendency_compute()`, respectively, and renames `marbl_comp_nhx_surface_emis()` to `marbl_nhx_surface_emis_compute()`.

Addresses #86


mnlevy1981:
Still to do: clean up the module-level comments in all these new modules. The original comments from `marbl_mod.F90` ended up in `marbl_surface_flux_mod.F90`, and the comments from `marbl_ciso_mod.F90` are in `marbl_ciso_surface_flux_mod.F90` . It's not entirely clear to me where documentation like the following belongs:

```
  !  Multispecies ecosystem based on Doney et al. 1996, Moore et al., 2002
  !  Based on POP Global NCAR Nitrogen Ecosystem Model
  !  version 0.0 (June 15th, 1998) from S.C. Doney.
  !  Based on Doney et al., 1996 model.
  !  Climate and Global Dynamics, NCAR
  !  (doney@whoi.edu)
  !
  !  Version 1.0
  !  Multispecies, multiple limiting nutrient version of ecosystem
  !  based on mixed layer model of Moore et al.(2002).  Implemented here with
  !  fixed elemental ratios and including only the diatoms and small
  !  phytoplankton, with a parameterization of calcification,
  !  by Keith Lindsay and Keith Moore, Fall 2001 - Spring 2002.
  !  Calcification parameterization based on Moore et al. 2002.
  !
  !  Version 2.0, January 2003
  !    Adds diazotrophs as a phytoplankton group, (based on Moore et al., 2002a)
  !    Allows for variable fe/C for all phytoplankton groups
  !     Allows for variable si/C for the diatoms
  !     Adds explicit tracers for DON, DOP, DOFe
  !     variable remin length scale for detrital soft POM and bSi f(temperature)
  !     Extensive modifications to iron scavenging parameterization
  !     Addition of a sedimentary dissolved iron source,
  !        (implemented in ballast code as excess remin in bottom cell)
  !        coded by J.K. Moore, (jkmoore@uci.edu)
  !
  !   Version 2.01. March 2003
  !     corrected O2 bug
  !     corrected grazing parameter z_grz bug at depth
  !     dust dissolution at depth releases iron,
  !     increased length scale for dust diss., increased hard fraction dust
  !     no deep ocean reduction in scavenging rates,
  !     increase bSi OC/ballast ratio 0.3 -> 0.35,
  !     corrected bug in diazotroph photoadaptation, and diat and sp adapatation
  !
  !   Version 2.02.
  !     corrected bug in Fe_scavenge (units for dust), May 2003
  !     changed C/N/P ratios to 117/16/1 (Anderson & Sarmiento, 1994)
  !
  !   Version 2.03., July 2003
  !     Remin of DOM no longer temperature dependent,
  !     new iron scavenging parameterization added,
  !     some dissolution of hard fraction of ballast materials added
  !
  !   Version 2.1, September 2003
  !     modfied iron scavenging and dust dissolution at depth
  !
  !   Version 2.11, March 2004
  !     fixed bug in iron scavenging code, replace dust and POC flux_in w/ flux_out
  !
  !   Version 2.12, April 2004 - Final version for GBC paper revision,
  !     (Questions/comments, Keith Moore - jkmoore@uci.edu
  !
  !   References
  !   Doney, S.C., Glover, D.M., Najjar, R.G., 1996. A new coupled, one-dimensional
  !   biological-physical model for the upper ocean: applications to the JGOFS
  !   Bermuda Time-Series Study (BATS) site. Deep-Sea Res. II, 43: 591-624.
  !
  !   Moore, JK, Doney, SC, Kleypas, JA, Glover, DM, Fung, IY, 2002. An intermediate
  !   complexity marine ecosystem model for the global domain. Deep-Sea Res. II, 49:
  !   403-462.
  !
  !   Moore, JK, Doney, SC, Glover, DM, Fung, IY, 2002. Iron cycling and nutrient
  !   limitation patterns in surface waters of the world ocean. Deep-Sea Res. II,
  !   49: 463-507.
```

Also, there are two things I want to look into doing:

1. Split `marbl_diagnostics_mod.F90` into `marbl_diagnostics_mod.F90` and `marbl_ciso_diagnostics_mod.F90`
2. Always call `marbl_ciso_*()` routines, but have `if (.not. ciso_on) return` at the top... i.e. move the `if (ciso_on)` checks into the Ciso routines. I think that will be a cleaner template to follow when adding additional tracer modules such as Niso.

mnlevy1981:
I've split `marbl_diagnostic_mod.F90`, and now realize that there are lots of variables, data types, and subroutines still using `surface_forcing` and `interior_forcing` -- I will be renaming all of those to `surface_flux` and `interior_tendency`, respectively.

mnlevy1981:
To clarify the last comment, I want to replace `surface_forcing` with `surface_flux` in relationship to the computation done in `marbl_surface_flux_compute()` (formerly `set_surface_forcing()`). Datatypes and subroutines dealing with forcing fields requested by MARBL (`u10_sqr`, `sst`, `Salinity`, etc) should keep `forcing` in its moniker, and we do separate those forcings based on whether they apply to the surface flux computation or the interior tendency computation. Maybe those should be referred to `forcings_for_surface_flux` and `forcings_for_interior_tendency`? That's starting to get pretty long, especially for things like `marbl_forcings_for_surface_flux_indexing_type` and `marbl_forcings_for_interior_tendency_indexing_type`

src/marbl_surface_flux_mod.F90
```diff
@@ -0,0 +1,567 @@
...
+         marbl_tracer_indices     = marbl_tracer_indices,     &
+         saved_state              = saved_state,              &
+         saved_state_ind          = saved_state_ind,          &
+         surface_flux_diags       = surface_flux_diags)
+
+    !-----------------------------------------------------------------------
+    ! Compute carbon isotopes surface fluxes
+    !-----------------------------------------------------------------------
+
+    ! pass in sections of surface_input_forcings instead of associated vars because of problems with intel/15.0.3
```
mnlevy1981: Having the `if (ciso_on)` check in `marbl_ciso_surface_flux_mod` means I'm always trying to pass `surface_input_forcings(ind%d1{3,4}c_id)`, and those indices are 0 for non-ciso runs. So I think we should pass all of `surface_input_forcings` (as well as `ind`) and let the subroutine pull out `sst`, `d13c`, and `d14c`.

Issue #317, opened 2018-09-07
Several PFT types have confusing names and dimensions


mnlevy1981:
1. Rename the following:

   * `autotroph_secondary_species_type` -> `autotroph_derived_terms_type`
   * `zooplankton_secondary_species_type` -> `zooplankton_derived_terms_type`
   * `autotroph_type` -> `autotroph_settings_type`
   * `zooplankton_type` -> `zooplankton_settings_type`

1. Move `(autotroph_cnt, km)` dimensions into `autotroph_local_type` and `autotroph_derived_terms_type` [same for `(zooplankton_cnt, km)` and the `zooplankton` equivalents]

   * Instead of declaring local variables inside `interior_tendency_compute()`, move variables into `marbl_interface_class` and construct them during `init`.
   * This should help with memory access as we implement #53

mnlevy1981:
To go along with the changes from `autotroph_type` -> `autotroph_settings_type` and `zooplankton_type` -> `zooplankton_settings_type`, I also changed `grazing_type` -> `grazer_settings_type`. This also led to variable name changes:

* `autotrophs` -> `autotroph_settings`
* `zooplankton` -> `zooplankton_settings`
* `grazing` -> `grazer_settings`

These changes propagate through the settings YAML files, MARBL python scripts, and POP wrappers to the MARBL python scripts (but don't change answers).

mnlevy1981:
Rather than `grazer_settings`, I ended up renaming `grazing` to `grazing_relationship_settings`. This issue is addressed in [marbl0.33.0](https://github.com/marbl-ecosys/MARBL/releases/tag/marbl0.33.0)

Issue #53, opened 2016-04-28
migrate k loop inside of subroutines called from marbl_set_interior_forcing


klindsay28:

mnlevy1981:
If this can be done in a bit-for-bit manner, it should be done before CESM 2.1 (in conjunction with #106).

Also, we should refactor several datatypes to be column-oriented (`autotroph_secondary_species` comes to mind). See also #237

mnlevy1981:
As of [marbl0.33.0](https://github.com/marbl-ecosys/MARBL/releases/tag/marbl0.33.0), this has been done except for the following loop that remains in `interior_tendency_compute()`:

```
    do k = 1, km

       call compute_scavenging(k, km, marbl_tracer_indices, tracer_local(:,:), &
            POC, P_CaCO3, P_SiO2, dust, Fefree(:), Fe_scavenge_rate(:), &
            Fe_scavenge(:), Lig_scavenge(:), marbl_status_log)

       if (marbl_status_log%labort_marbl) then
          call marbl_status_log%log_error_trace('compute_scavenging()', subname)
          return
       end if

       call compute_large_detritus_prod(k, domain, marbl_tracer_indices, zooplankton_derived_terms, &
            autotroph_derived_terms, Fe_scavenge(k),                    &
            POC, POP, P_CaCO3, P_CaCO3_ALT_CO2, P_SiO2, dust, P_iron,   &
            dissolved_organic_matter%DOP_loss_P_bal(k), marbl_status_log)

       ! FIXME #28: need to pull particulate share out
       !            of compute_particulate_terms!
       call compute_particulate_terms(k, domain,                   &
            marbl_particulate_share, p_remin_scalef(k),            &
            POC, POP, P_CaCO3, P_CaCO3_ALT_CO2,                    &
            P_SiO2, dust, P_iron, PON_remin(k), PON_sed_loss(k),   &
            QA_dust_def(k),                                        &
            tracer_local(:, k), carbonate, sed_denitrif(k),        &
            other_remin(k), fesedflux(k), marbl_tracer_indices,    &
            glo_avg_fields_interior_tendency, marbl_status_log)

       if (marbl_status_log%labort_marbl) then
          call marbl_status_log%log_error_trace('compute_particulate_terms()', subname)
          return
       end if

       if  (k < km) then
          call update_particulate_terms_from_prior_level(k+1, POC, POP, P_CaCO3, &
[...]

Issue #158, opened 2017-05-09
rename subroutine marbl_compute_autotroph_phyto_diatoms


jessluo:
The current subroutine name is not informative. This subroutine appears to modify the nutrient flux terms for the different autotroph groups. How about `marbl_compute_autotroph_nutrient_flux` ?

mnlevy1981:
This was fixed in #318 but I forgot to close the issue

PR #318, opened 2018-09-07
Clean up PFT-related data structures and migrate k-loop out of interior_tendency_compute()

mnlevy1981:
Many data structures had names that were not descriptive, so it was confusing what the purpose of the structures were. Also, there were several datatypes that were declared as arrays when it made more sense to have the array dimensions inside the datatype.

These changes go hand-in-hand with moving the `k` loop out of `interior_tendency_compute()` -- moving the array dimensions inside the type will make it easier to vectorize computations across a single column rather than needing to loop over each level.

This pull request is restricted to being bit-for-bit, so there will likely be additional vector-based optimizations to make at a later date.

Addresses #53 (perhaps not entirely) and #317. The Fortran interface will not change, but the names of variables in the settings file will - that will affect the GCMs and also require documentation updates.

### DIFF
--- a/MARBL_tools/MARBL_settings_file_class.py
+++ b/MARBL_tools/MARBL_settings_file_class.py
@@ -228,11 +228,7 @@ class MARBL_settings_class(object):
         append_to_keys = (('PFT_defaults == "CESM2"' in self._config_keyword) and
                           (category_name == "PFT_derived_types"))
         if append_to_keys:
-            PFT_name = variable_name[:-9] # drop '_settings' from 'autotroph_settings', 'zooplankton_settings', or 'grazer_settings'
-            if PFT_name == 'zooplankton':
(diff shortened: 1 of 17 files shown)


mnlevy1981:
This pull request is not ready to be merged yet, but I want

1. to start checking the Travis CI test results
1. leaving PR-specific comments on work that needs to be done here instead of spread across the issue tickets being addressed

mnlevy1981:
Address #158 with `compute_autotroph_nutrient_uptake()` when I get to that subroutine

mnlevy1981:
6a4b02e is bit-for-bit with `aux_pop_MARBL` baselines on hobart; NLCOMP failures (as expected) because `grazer_settings` was renamed `grazing_relationship_settings` in `marbl_in`. I will continue down my list of comments from the code review, but the goal is to launch `aux_pop` tests on cheyenne at the end of the day and make new issue tickets for whatever clean-up I don't get to. (I'll update documentation after the POP tag is made.)

PR #327, opened 2018-11-15
Remove workaround to PGI compiler bug

mnlevy1981:
As of 18.10, put_input_file_line() does not need the pgi_bugfix_var optional
argument to be recognized as a unique interface for put_setting(). Once this is
accepted onto development, PGI 18.4 (and earlier) will not be able to compile
MARBL. It will fail to build the stand-alone driver with the error

```
PGF90-S-0155-cannot access PRIVATE type bound procedure put_input_file_line
```

(Presumably GCMs that use put_setting() will also run into this error)

### DIFF
--- a/src/marbl_interface.F90
+++ b/src/marbl_interface.F90
@@ -541,7 +541,7 @@ contains
 
   !***********************************************************************
 
-  subroutine put_input_file_line(this, line, pgi_bugfix_var)
+  subroutine put_input_file_line(this, line)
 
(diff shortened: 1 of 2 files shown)


mnlevy1981:
I've asked ISG to install the newest PGI compiler on hobart, at which point I'll make a CIME pull request to update the version used by CESM... after CESM is building with 18.10, we can bring this onto development and stable (assuming the CIME change goes into both CESM 2.1 and 2.2)

Issue #231, opened 2018-02-22
Make additional ecosystem parameters available via marbl_settings_mod or namelist


jessluo:
A few parameters are encoded in the source files and should be made available for modification via `marbl_settings_mod.F90` or as namelist parameters:

1. Power of the 'quadratic' loss term for autotrophs and zooplankton. For autotrophs, the value is 1.75 (e.g. `Pprime(auto_ind)**1.75_r8`) and for zooplankton the value is 1.5 (e.g. `Zprime(zoo_ind)**1.5_r8`). 
It may not be necessary to allow for individual variations in the 'power loss power' term. One value for autotrophs and one value for zooplankton may be sufficient.

2. Power of the 'quadratic' term on the implicit calcifier routing to POC. Within the `marbl_compute_grazing` subroutine, the routing of autotroph grazing (of implicit calcifiers) to POC has a power of 1.6. Should that be a parameter within marbl_settings_mod as well? Also, should this portion of the subroutine be more general to include autotrophs that are neither implicit calcifiers nor diatoms?

3. Maximum CaCO3 : C ratio (`QCaCO3_max`) should be flexible by autotroph PFT. Currently, since there is only one implicit calcifier in the default CESM2 parameter set, there is no need for variations in QCaCO3_max. However, under cases where there is more than 1 implicit calcifier, it would be nice to adjust the maximum CaCO3 : C ratio per implicit calcifier.

4. Similarly, should `f_photosp_CaCO3` be changed as well to allow for variations between implicit calcifiers (if implicit calcifier > 1)? Right now it appears that `f_photosp_CaCO3` is the same value as `QCaCO3_max`, so if `QCaCO3_max` changes, then `f_photosp_CaCO3` should as well.

mnlevy1981:
For point 4, `f_graze_sp_poc_lim` and `spc_poc_fac` should also be moved to `autotroph_type` (and it's not entirely clear that "if `QCaCO3_max` changes, then `f_photosp_CaCO3` should as well" is true, need to look into it)

matt-long:
I think we should remove `sp` from the names of parameters associated with implicit calcification and replace with something like `imp_calcifier` or `imp_calc`.  Additionally, the following parameters might be renamed to be more informative.
- `parm_f_prod_sp_CaCO3` is really a minimum or base rate, so could be `parm_f_imp_calc_photo_CaCO3_min`
[...]

PR #328, opened 2018-11-20
Further enabling size-resolved plankton model

jessluo:
A set of changes to enable the Size-structured Plankton Ecological Traits (SPECTRA) model, which is a size-resolved plankton model, with bit-for-bit reproduction of the current stable release of MARBL.

- Introduction of an alternate temperature scaling function, based on the Arrhenius temperature function
  - MARBL namelist flag `l_ea_on` turns on or off the alternate temperature scaling, with default values of `Ea = 0.32` for phytoplankton and `Ea = 0.65` for zooplankton.
- Diagnostics changes to enable outputs of per predator-prey grazing fluxes
  - `graze_{auto}_zoo` is replaced by `graze_{auto}_zootot`
  - `graze_{auto}_{zoo}` terms are added


mnlevy1981:
We'll get this onto `development` soon (probably after the holidays), but I'd like to include a fix for #231 on this branch. I'm happy to help with some of the coding - we can talk about the specifics later.

mnlevy1981:
@jessluo -- on Friday we talked about how I was seeing round-off level answer changes in the CESM test suite. With @klindsay28's help, we realized that

```
-        zoo_loss(zoo_ind,:) = (zooplankton_settings(zoo_ind)%z_mort2_0 * Zprime(zoo_ind,:)**1.5_r8 &
+        zoo_loss(zoo_ind,:) = (zooplankton_settings(zoo_ind)%z_mort2_0 * Zprime(zoo_ind,:)**zoo_mort2_exp &
                                + zooplankton_settings(zoo_ind)%z_mort_0  * Zprime(zoo_ind,:)) * Tfunc_zoo(zoo_ind,:)
```

is changing answers because of the way the intel compiler treats `**1.5_r8` differently than `**n` for a variable `n`. (One theory is that `x**1.5` can be compiled as `x * sqrt(x)` while `x**zoo_mort2_exp` can't be simplified at compile time; it's definitely something along these lines.) So I'd like to back out just the `zoo_mort2_exp` parameter for this pull request. There are a few options for proceeding:

1. Immediately make a branch off of `development` with `zoo_mort2_exp` as a parameter, and merge it in once we are ready for answer-changing commits (we have at least one update in the pipeline for the CESM 2.1 CMIP6 runs that need to be bit-for-bit)
1. Same as above, but after updating `development` and `stable` with this PR we could make a CESM 2.1 release branch for the bit-for-bit work and then immediately bring in the `zoo_mort2_exp` update
1. Add a kludge to the code that checks to see if `zoo_mort2_exp == 1.5`:
```
if (zoo_mort2_exp .eq. 1.5_r8) then
        zoo_loss(zoo_ind,:) = (zooplankton_settings(zoo_ind)%z_mort2_0 * Zprime(zoo_ind,:)**1.5_r8 &
                                + zooplankton_settings(zoo_ind)%z_mort_0  * Zprime(zoo_ind,:)) * Tfunc_zoo(zoo_ind,:)
else
        zoo_loss(zoo_ind,:) = (zooplankton_settings(zoo_ind)%z_mort2_0 * Zprime(zoo_ind,:)**zoo_mort2_exp &
                                + zooplankton_settings(zoo_ind)%z_mort_0  * Zprime(zoo_ind,:)) * Tfunc_zoo(zoo_ind,:)
end if
```
It's not pretty to look at (and I may strip it out after the final CESM 2.1 release tag), but it probably would give us a slight performance improvement so it wouldn't *just* be a kludge for the sake of not changing answers.

Do you have any preference? I'm leaning towards (1) -- adding `zoo_mort2_exp` back in is a small change, so it should be any easy fix when we are ready. If you want to run more tests with `zoo_mort2_exp = 2.0` you can run off the branch (and I can keep it up-to-date with `development`, e.g. after bringing in the coccolithophore branch)

jessluo:
My vote is for the first option that @mnlevy1981 mentioned, but with a modification. `zoo_mort2_exp` got added as a settings variable at the same time as `auto_mort2_exp`, so it would probably make sense to keep both `zoo_mort2_exp` and `auto_mort2_exp` out of this pull request until we are able to make answer changing modifications to the code base.

klindsay28:
@mnlevy1981, it seems to me that your option 1, and thus also 2, will change answers.
So I think that I'm misunderstanding what you are proposing.

Perhaps I'm getting tripped up by terminology. Does the word 'parameter' mean the same thing in the following sentences from your comment?

> So I'd like to back out just the `zoo_mort2_exp` parameter for this pull request.

> 1) Immediately make a branch off of development with `zoo_mort2_exp` as a parameter, ...

mnlevy1981:
@klindsay28 I'm using parameter in both sentences to mean "MARBL settings variable" rather than "Fortran parameter." So backing out the changes implies removing the variable named `zoo_mort2_exp` (introduced in 539ecb9) and keeping the hard-coded `**1.5_r8` in this PR. Accepting this PR would then be bit-for-bit with the existing `development` and `stable` branches.

Both (1) and (2) will change answers, but my intent in (1) is to have those answer changes on a branch (perhaps named `mort2_exp_in_settings`) so that `development` and `stable` remain bit-for-bit. In (2) I would make a `cesm2.1` release branch at the current head of (read: after accepting this PR) `development` & `stable` before bringing `zoo_mort2_exp` to `development`, so we would change answers on our main branches but have a way to keep CESM 2.1 tags bit-for-bit.

Per @jessluo's last comment, I can also remove `auto_mort2_exp` (keeping `**1.75_r8` in this PR) and then adding the `auto_mort2_exp` variable when we are ready for `zoo_mort2_exp` and the answer changes it brings.

mnlevy1981 (approved):
After talking with @klindsay28 and @matt-long, we decided to go with a fourth option: I'll start the `cesm2.1` release branch at the current head of `stable` and then bring this onto `development` as part of the CESM 2.2 development (it will not be available in CESM 2.1 at all).

PR #330, opened 2019-02-22
MARBL version with explicit calcifiers

kristenkrumhardt:
Version of MARBL with an explicit calcifier phytoplankton functional type, based on coccolithophores; the parameterization is described in Krumhardt et al. (JAMES) .

Carbon fixation (photosynthesis) of the coccolithophore PFT sensitive to aqueous CO2. CO2 is handled as an additional nutrient required for photosynthesis.

Calcification by the coccolithophore PFT is sensitive to temperature, aqueous CO4, and PO4 limitation.


src/marbl_settings_mod.F90
```diff
@@ -134,7 +134,7 @@ module marbl_settings_mod
       f_graze_sp_poc_lim    = 0.36_r8,  &
       f_photosp_CaCO3       = 0.40_r8,  & ! proportionality between small phyto
                                           !    production and CaCO3 production
-      f_graze_CaCO3_remin   = 0.33_r8,  & ! fraction of spCaCO3 grazing which is remin
```
mnlevy1981: I assume this line will change answers in the default POP setup? Maybe we should move `f_graze_CaCO3_remin` to the settings file with a default of 0.33 and then part of running with coccolithophores would be changing the value to 0.7 in `user_nl_marbl`

kristenkrumhardt: Yes, I think that would be a good idea.

mnlevy1981:
Testing update:

Ahead of helping @kristenkrumhardt make some changes following the code review earlier this month, I ran `aux_pop_MARBL` on hobart. Mostly positive results, and we already have plans to address the bad results:

1. This branch runs in debug mode (and with the nag compiler) -- previous tests were intel-only and without debug
1. Namelist comparisons fail due to adding variables to `user_nl_marbl`, but this was expected
1. The addition of two new history fields is causing minor testing hiccups, but addressing kristenkrumhardt/MARBL#2 will fix both of these failures:

   1. `cprnc` now triggers a failure when a variable is in one file but not the other (this will go away because we will not be adding `C_lim_surf` and `C_lim_Cweight_avg_100m` for `sp`)
      ```
      Variable on file1: sp_C_lim_surf not found on file2
      Variable on file1: sp_C_lim_Cweight_avg_100m not found on file2
      ```
   1. exact restart tests are failing because these two fields are changing, but again this will go away because these two fields shouldn't be computed in the first place
      ```
      RMS sp_C_lim_surf                    1.3772E+14            NORMALIZED  8.7393E+00
      RMS sp_C_lim_Cweight_avg_100m        2.3853E+13            NORMALIZED  8.1829E+00
      ```

PR #338, opened 2019-07-19
New stand-alone test that calls compute() functions

### LINKED ISSUE #156: Design stand-alone test that calls set_surface_forcing and set_interior_forcing
Testing for many issues (#41 and #146 stand out) would be easier if we had a simple test that initialized the model, called `marbl_instance%set_surface_forcing` and `marbl_instance%set_interior_forcing`, and then wrote the diagnostic output.

This would require a couple of things that we don't have right now:

1. Datasets (initial conditions, forcing files, etc), and a way to provide them to users (some sort of inputdata repository?)
1. An IO module that can parse the diagnostics type and write netCDF output (CVMix has one that would be a great starting point)

mnlevy1981:
Create the `compute_cols` test which calls `surface_flux_compute()` and `interior_tendency_compute()`. This required lots of infrastructure updates:

1. Created `marbl_io_mod.F90` to handle reading / writing netcdf files (need to read in initial conditions and forcing fields, and also write marbl diagnostic fields)
1. Updated the build system updated to support linking with netcdf library
1. Created a location in the repository to store the initial conditions and forcing fields
1. Created a location in the repository to store baselines for regression testing
1. Created a tool to compare different netcdf files (ensure that files are bit-for-bit regardless of number of instances in run, and also ensure that output is "close" to baselines for some value of close)
1. Added a way to set the default input file a test is run with

I also had the opportunity to clean up a lot of code along the way:

* Some of the `intent` attributes in the driver were misleading (`inout` when `out` would suffice)
* The `lecovars_full_depth_tavg` flag now applies to all 3D diagnostics, not just the tracer metadata (and the carbon isotope module uses it instead of having its own `ciso_lecovars_full_depth_tavg`
* Removed `yellowstone` configuration for the stand-alone test

I set up a [kanban board](https://github.com/mnlevy1981/MARBL/projects/1) to track progress on this - although the tests pass, it's not quite ready to be merged. I'll leave a comment when this is ready for review.


mnlevy1981:
The Fortran code (as well as the auxiliary python scripts, YAML files, etc) is finished - it has been through three rounds of extensive review and is ready to be merged. I still need to update the documentation and include instructions for generating the initial conditions and forcings for the new test, so it's not quite time to merge these changes.

mnlevy1981:
I'm hosting the documentation for my branch on [read the docs](https://marbl-levy.readthedocs.io/en/enhancement-test_set_interior/). I've moved the POP examples into the [implementation](https://marbl-levy.readthedocs.io/en/enhancement-test_set_interior/implementations/pop/index.html) section, and now the [How to Use MARBL in a GCM](https://marbl-levy.readthedocs.io/en/enhancement-test_set_interior/usr-guide/GCM-interface/index.html) section contains examples from the stand-alone driver.

I think this is a good start for the documentation, but I wouldn't be surprised if someone else reads through it and notices some important content is missing.

PR #344, opened 2020-05-07
Fix Travis CI

### LINKED ISSUE #343: Travis CI issues?
First, it looks like test results from #342 were never actually reported in the PR. (I think I fixed this by revoking access to Travis CI and then reinstating it... will find out for sure with next PR).

Even worse, though, it looks like the tests actually [failed](https://travis-ci.org/github/marbl-ecosys/MARBL/builds/684377969). Python 2.7 wasn't able to get past the `pip install` and Python 3.6 was throwing exceptions in the netCDF comparison:

```
Traceback (most recent call last):
  File "./netcdf_comparison.py", line 274, in <module>
    ds_base_in, ds_new_in = _open_files(args.baseline, args.new_file) # pylint: disable=invalid-name
  File "./netcdf_comparison.py", line 66, in _open_files
    return xr.open_dataset(baseline), xr.open_dataset(new_file)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/xarray/backends/api.py", line 499, in open_dataset
    engine = _get_default_engine(filename_or_obj, allow_remote=True)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/xarray/backends/api.py", line 153, in _get_default_engine
    engine = _get_default_engine_netcdf()
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/xarray/backends/api.py", line 96, in _get_default_engine_netcdf
    import netCDF4  # noqa: F401
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/netCDF4/__init__.py", line 3, in <module>
    from ._netCDF4 import *
  File "netCDF4/_netCDF4.pyx", line 1213, in init netCDF4._netCDF4
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/cftime/__init__.py", line 1, in <module>
    from ._cftime import utime, JulianDayFromDate, DateFromJulianDay
  File "__init__.pxd", line 918, in init cftime._cftime
ValueError: numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject
```

The good news is that the failures all appear to be in the Travis environment, rather than indicative of actual issues with the code. I bet a fix for this is to pin versions of packages in the pip call.

A few related issues:

1. This isn't the first time I've had TravisCI stop testing a repository (possibly not even the first time it stopped testing **this** repository) and it's a little frustrating. Perhaps CircleCI would be more stable?
1. I don't know if we need to continue testing with Python 2.7; I'm happy to continue doing so until we've had a chance to discuss it, but I don't think anyone is still relying on python 2 at this point
1. I'd like to add python 3.7 (and maybe 3.8?) to the test suite. I don't expect any failures from those later versions, but if there ARE issues it would be nice to find out from continuous integration rather than surprising a user with an incompatibility

mnlevy1981:
Need to ensure that Travis results appear in the pull request and that Travis completes successfully. At that point, I may add python 3.7 and 3.8 testing.

Fixes #343

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,10 @@ addons:
          - netcdf-bin libnetcdf-dev openmpi-bin libopenmpi-dev gfortran
 
 install:
-   - pip install pyyaml numpy "xarray<=0.15.0" netcdf4 "cftime<1.1.2"
+   - pip install pyyaml "numpy>=1.16" xarray netcdf4
    - pip install -r docs/py_requirements.txt
 
 python:
-   - '2.7'
    - '3.6'
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,8 @@ install:
 
 python:
    - '3.6'
+   - '3.7'
+   - '3.8'
 
 script:
    - cd docs/src; make html
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,6 @@ install:
 
 python:
    - '3.6'
-   - '3.7'
-   - '3.8'
 
 script:
    - cd docs/src; make html
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ addons:
          - netcdf-bin libnetcdf-dev openmpi-bin libopenmpi-dev gfortran
 
 install:
-   - pip install pyyaml numpy xarray netcdf4
+   - pip install pyyaml numpy xarray netcdf4 "cftime<1.1.2"
    - pip install -r docs/py_requirements.txt
 
 python:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ addons:
          - netcdf-bin libnetcdf-dev openmpi-bin libopenmpi-dev gfortran
 
 install:
-   - pip install pyyaml numpy xarray netcdf4 "cftime<1.1.2"
+   - pip install pyyaml numpy "xarray<=0.15.0" netcdf4 "cftime<1.1.2"
    - pip install -r docs/py_requirements.txt
 
 python:


mnlevy1981:
Travis is launching, which is promising! 🤞 that tests pass rather than reporting environment issues

mnlevy1981:
Having trouble with 3.7 and 3.8:

```
Downloading archive: https://storage.googleapis.com/travis-ci-language-archives/python/binaries/ubuntu/14.04/x86_64/python-3.7.tar.bz2
$ curl -sSf --retry 5 -o python-3.7.tar.bz2 ${archive_url}
curl: (22) The requested URL returned error: 404 Not Found
Unable to download 3.7 archive. The archive may not exist. Please consider a different version.
```

```
Downloading archive: https://storage.googleapis.com/travis-ci-language-archives/python/binaries/ubuntu/14.04/x86_64/python-3.8.tar.bz2
$ curl -sSf --retry 5 -o python-3.8.tar.bz2 ${archive_url}
curl: (22) The requested URL returned error: 404 Not Found
Unable to download 3.8 archive. The archive may not exist. Please consider a different version.
```

So I guess this means relying on `conda` instead of `pip` to build our test environment? I'm going to revert the last commit and open a new issue re: `conda` (which might get us 2.7 testing again)... but for the moment, the only Travis testing will rely on python 3.6.

mnlevy1981:
@klindsay28 -- let me know if you want to chat about this PR; I'm thinking this is just a placeholder to get some automated testing back online but I don't really have a timeline for when I'd try to come up with a more permanent solution... ideally we'd only limp along with this testing for a couple of weeks before moving to something more robust.

klindsay28 (approved):

Issue #341, opened 2020-05-06
Create a tool that generates a table of settings values


mnlevy1981:
@matt-long asked if there was an easy way to produce a table of default settings values for a paper (so output would be in LaTeX) and also for the web documentation (so reStructuredText or maybe just re-use the LaTeX?). I think the best approach would be a new python script that would populate a `pandas` dataframe and use the `to_latex()` function to produce the desired output; input would be a YAML file with the following information

* Requested columns in the table (match keys in the settings YAML file)
* Output format
* List of variables to include in the table
* For each variable, a way to override defaults for things like variable name, `longname`, `units`, etc (to allow for better formatting)

We can take advantage of the fact that `MARBL_tools/MARBL_generate_settings_file.py` already does a lot of the heavy-lifting we need to make sure we get the right default value for each variable.

mnlevy1981:
Closing this -- #346 is the only update the belongs in this repo, and an example of how to use it can be found in https://github.com/marbl-ecosys/cesm2-marbl/pull/37

PR #346, opened 2020-05-08
Update settings_dict to hold more data

mnlevy1981:
rather than `settings_dict[varname]` being set to the value of the variable, it
is now a dictionary with `value`, `units`, and `longname` keys. This is being done to
make it easier to turn the YAML file into a table to publish either in a
journal article or online through sphinx.

I don't think POP uses `settings_dict` directly, but this may require a small POP update as well.


mnlevy1981:
I tried using my branch in the latest version of POP and needed to modify `MARBL_scripts/MARBL_wrappers/MARBL_settings.py`; so POP _will_ need some updates to use MARBL after we make this change.

****

I did want to mention that there is a little bit more data in settings file that we could consider bringing in to `settings_dict`:

```
   parm_o2_min :
      longname : Minimum O2 needed for production & consumption
      subcategory : 4. general parameters
      units : nmol/cm^3
      datatype : real
      default_value : 5.0
```

The `subcategory` is really just used for some internal bookkeeping in MARBL (and to make sure values are determined in the proper order in cases where one variable depends on another), but maybe I should include `datatype` as well? I could see that being useful information in the web documentation.

mnlevy1981:
@klindsay28 -- I'd like to merge this change in, but don't want to do so without a review. I think it might be small enough to look at asynchronously, but I'm happy to walk through it with you if you'd prefer (my calendar is up-to-date). The impetus for this change is that https://github.com/marbl-ecosys/cesm2-marbl/pull/37 pulls more than just the value of a parameter out of `settings_dict`... so before `settings_dict[varname]` returned the parameter value but now it's a dictionary containing the long name and units in addition to the value itself. This doesn't fix #341 but the PR in [marbl-ecosys/cesm2-marbl](https://github.com/marbl-ecosys/cesm2-marbl) will get us closer to addressing that issue.

Once this is approved and merged, I'll make a POP PR that adds `['value']` to all seven references to `self._MARBL_settings.settings_dict` in `MARBL_scripts/MARBL_wrappers/MARBL_settings.py`. I have a sandbox with all these changes made and have verified that all `aux_pop_MARBL` tests on cheyenne pass both `NLCOMP` and `BASELINE`, so `marbl_in` does not change relative to baselines and the history files are bit-for-bit.

klindsay28:
Would it be simpler for `self._settings_dict_attrs` to be a local variable in `_update_settings_dict`? It's not used outside of this function.

mnlevy1981:
> Would it be simpler for `self._settings_dict_attrs` to be a local variable in `_update_settings_dict`? It's not used outside of this function.

yes it would! For some reason I thought there were a couple of routines that would need it, and when I noticed my changes were all in `_update_settings_dict()` I didn't think about moving it. Good catch!

klindsay28 (approved):
looks good

PR #372, opened 2021-02-25
Initialize more arrays below kmt

### LINKED ISSUE #371: Make sure all arrays are initialized before use
@mvertens reported an error while testing POP's NUOPC driver that manifest itself as float point errors in `compute_scavenge()` that she traced back to uninitialized values in `dust%hflux_in`. This seemed very similar to what @klindsay28 saw [back in 2018](https://github.com/marbl-ecosys/MARBL/issues/176#issuecomment-374387403). Taking my own advice from back then, I ran the `call_compute_subroutines` with the intel compiler using the `-g -O0 -init=snan,arrays` compile flags (among others). Sure enough, the tracebacks were pointing to the same line of `compute_scavenge()` that was causing problems in the NUOPC tests.

mnlevy1981:
Running call_compute_subroutines with the intel compiler using the following
FFLAGS: `-g -O0 -init=snan,arrays` crashed until I initialized several fields
below the sea floor.

Leaving this is draft status until @mvertens verifies that it fixes the error she was seeing.

Fixes #371

### DIFF
--- a/src/marbl_ciso_interior_tendency_mod.F90
+++ b/src/marbl_ciso_interior_tendency_mod.F90
@@ -1074,6 +1074,7 @@ contains
     use marbl_settings_mod , only : caco3_bury_thres_iopt_fixed_depth
     use marbl_settings_mod , only : caco3_bury_thres_depth
     use marbl_settings_mod , only : caco3_bury_thres_omega_calc
+    use marbl_interior_tendency_share_mod, only : marbl_interior_tendency_share_set_used_particle_terms_to_zero
 
     integer (int_kind),                       intent(in)    :: k                 ! vertical model level
     type(marbl_domain_type),                  intent(in)    :: domain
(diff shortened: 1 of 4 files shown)


mnlevy1981:
It would be useful to figure out what `gfortran` flags can be set in order to detect this error, then our continuous integration can ensure similar problems don't slip back in.

mnlevy1981:
Also, there might be similar problems with the ciso code but we don't have a test for that

mnlevy1981:
Verified this is sufficient to get @mvertens's test to pass

mnlevy1981:
@klindsay28 as it turns out, the `-finit-real=snan` flag isn't sufficient to trap unitialized real variables in a derived type. I added `-finit-derived` per [some gfortran documentation](https://gcc.gnu.org/onlinedocs/gfortran/Code-Gen-Options.html), except it includes the note

> These options do not initialize
> 
> * objects with the POINTER attribute
> * allocatable arrays
> * variables that appear in an EQUIVALENCE statement. 

And these are allocatable arrays. So when cheyenne is back up I'll add `-init=snan,arrays` to the intel flags, verify it's aborting when I don't initialize `hflux_out` and `sflux_out`, and then use it to see if I also need to initialize `remin` (currently `marbl_interior_tendency_share_set_used_particle_terms_to_zero()` does not set `sinking_particle%remin(k) = c0` because I was too trusting of the gfortran flags, but I suspect I'll need to add that line back in). Then I'll test in CESM and, assuming all tests pass, merge this to development.

mnlevy1981:
Per the previous comment, I tried running the stand-alone test with just `hflux_out` and `sflux_out` set to 0 below the sea floor, and the intel compiler didn't like that. So dd7cb41 does two things:

1. sets `remin` to 0 below the sea floor
2. updates the compile flags for running the stand-alone driver with intel

I tested that commit on cheyenne by running `aux_pop_MARBL` with both intel and gnu -- both sets of tests were bit-for-bit identical with baselines, so I'll merge this in and make a new tag.

PR #377, opened 2021-04-30
Return bottom flux

mnlevy1981:
Create a new array `bottom_fluxes` in the `marbl_interface_class` that represents per-tracer flux into the bottom of the column.

Opening this as a draft PR so I can see CI test results (and better share code snippets), but it's not ready for review / merging yet.

Fixes #360

PR #381, opened 2021-07-06
Distribute particulate bottom flux within column

### LINKED ISSUE #360: change interface so that interior_tendency_compute returns vertical tracer flux at sea floor
`interior_tendency_compute` currently computes vertical tracer flux at the sea floor and adds the effect of this to `interior_tendencies`, in the 'bottom' model layer. This will be problematic in MOM6 when there are multiple vanishingly thin layers at the sea floor, as MARBL will generate very large tendencies in the bottom layer.

If MARBL returns this vertical tracer flux, then the GCM can apply it as is appropriate for the GCM. For GCM=MOM6, the flux can be an argument to MOM6's `tracer_vertdiff` subroutine. This will handle the presence of thin layers at the sea floor. For GCM=POP, the flux can be converted into a tendency in the model's bottom layer, as MARBL currently does. This will probably introduce roundoff differences in POP/MARBL solutions.

Note that the computation of some elemental conservation diagnostics (`Jint_{C,N,P,Si,Fe}tot`) will need to be updated to include these sea floor fluxes.

mnlevy1981:
This offers a different strategy to the same problem #377 is attempting to solve. Here, the particulate that hits the sea floor is redistributed as remineralization within the entire column (rather than just in the bottom layer, which is how `development` currently handles it, and without computing a bottom flux for each tracer, which is how the other PR handles it).

This is a draft because two key pieces are missing:

1. Providing a mechanism for the GCM to pass `bot_flux_to_tend(:)` instead of assuming it is `1 / delta_z(column_kmt)` at level `column_kmt` and 0 elsewhere
2. Updating the CISO driver to behave similarly

Note that this approach should be bit-for-bit in POP (which is not the case with the other PR due to how we handle denitrification) and does not require changes to the conservation diagnostics

Fixes #360

### DIFF
--- a/defaults/json/settings_cesm2.0.json
+++ b/defaults/json/settings_cesm2.0.json
@@ -888,6 +888,13 @@
          "subcategory": "4. general parameters (bury coeffs)",
          "units": "unitless"
       },
+      "bftt_dz_sum_thres": {
+         "datatype": "real",
+         "default_value": 1e-14,
+         "longname": "MARBL will abort if abs(1 - sum(bot_flux_to_tend)) exceeds this threshold",
+         "subcategory": "4. general parameters",
+         "units": "unitless"
+      },
       "bury_coeff_rmean_timescale_years": {
          "datatype": "real",
          "default_value": 10,
--- a/defaults/json/settings_cesm2.1+cocco.json
+++ b/defaults/json/settings_cesm2.1+cocco.json
@@ -930,6 +930,13 @@
          "subcategory": "4. general parameters (bury coeffs)",
          "units": "unitless"
       },
+      "bftt_dz_sum_thres": {
+         "datatype": "real",
+         "default_value": 1e-14,
+         "longname": "MARBL will abort if abs(1 - sum(bot_flux_to_tend)) exceeds this threshold",
+         "subcategory": "4. general parameters",
+         "units": "unitless"
+      },
       "bury_coeff_rmean_timescale_years": {
          "datatype": "real",
          "default_value": 10,
--- a/defaults/json/settings_cesm2.1.json
+++ b/defaults/json/settings_cesm2.1.json
@@ -888,6 +888,13 @@
          "subcategory": "4. general parameters (bury coeffs)",
          "units": "unitless"
       },
+      "bftt_dz_sum_thres": {
+         "datatype": "real",
+         "default_value": 1e-14,
+         "longname": "MARBL will abort if abs(1 - sum(bot_flux_to_tend)) exceeds this threshold",
+         "subcategory": "4. general parameters",
+         "units": "unitless"
+      },
       "bury_coeff_rmean_timescale_years": {
          "datatype": "real",
          "default_value": 10,
--- a/defaults/json/settings_latest+cocco.json
+++ b/defaults/json/settings_latest+cocco.json
@@ -930,6 +930,13 @@
          "subcategory": "4. general parameters (bury coeffs)",
          "units": "unitless"
       },
+      "bftt_dz_sum_thres": {
+         "datatype": "real",
+         "default_value": 1e-14,
+         "longname": "MARBL will abort if abs(1 - sum(bot_flux_to_tend)) exceeds this threshold",
+         "subcategory": "4. general parameters",
+         "units": "unitless"
+      },
       "bury_coeff_rmean_timescale_years": {
          "datatype": "real",
          "default_value": 10,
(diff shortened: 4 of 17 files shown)


mnlevy1981:
As of 816c3e5, GCM sets `marbl_instance%bot_flux_to_tend(:)` before each call to `marbl_instance%interior_tendency_compute()` (and the CISO fix went in dc90124, which was tested in POP and is bit-for-bit with marbl0.39.3 tag)

klindsay28:
Do you know if we need the increase in `max_bracket_grow_it` with `MIX_BOUNDARY_TRACER_ALE=True`?

If we need to increase `max_bracket_grow_it`, I'd like to understand why, to avoid the increase papering over a problem without us knowing that.

mnlevy1981:
> Do you know if we need the increase in `max_bracket_grow_it` with `MIX_BOUNDARY_TRACER_ALE=True`?
> 
> If we need to increase `max_bracket_grow_it`, I'd like to understand why, to avoid the increase papering over a problem without us knowing that.

That's a good question. Yesterday I mentioned not being able to run on `izumi`, but this morning I woke up and realized I didn't include the updated version of `MOM_tracer_diabatic.F90` in my sandbox, so it's possible the code was hanging due to the parentheses bug in `tracer_vertdiff()` rather than memory or compiler issues... so I'm trying one more time to run on the CGD machine and if it works, I'll see if I can lower `max_bracket_grow_it` again.

mnlevy1981:
> this morning I woke up and realized I didn't include the updated version of `MOM_tracer_diabatic.F90` in my sandbox, so it's possible the code was hanging due to the parentheses bug in `tracer_vertdiff()` rather than memory or compiler issues...

I also forgot to copy all the inputdata files over, which might actually have been the problem... so I guess on Thursday I can try again?

mnlevy1981:
> > Do you know if we need the increase in `max_bracket_grow_it` with `MIX_BOUNDARY_TRACER_ALE=True`?
> > If we need to increase `max_bracket_grow_it`, I'd like to understand why, to avoid the increase papering over a problem without us knowing that.
> 
> That's a good question.

`max_bracket_grow_it = 3` died in `/glade/scratch/mlevy/C1850MOMECO.087/run` despite `MIX_BOUNDARY_TRACER_ALE=True`

```
WARNING from PE   509: (Task 509) Message from (lon, lat) (  49.333,   6.777). Level: 65
581:WARNING from PE   509: (Task 509) MARBL WARNING (marbl_co2calc_mod:drtsafe): (marbl_co2calc_mod:drtsafe) it = 4
581:WARNING from PE   509: (Task 509) MARBL WARNING (marbl_co2calc_mod:drtsafe): (marbl_co2calc_mod:drtsafe) x1,f =  0.9175434E-008-0.9062212E-004
581:WARNING from PE   509: (Task 509) MARBL WARNING (marbl_co2calc_mod:drtsafe): (marbl_co2calc_mod:drtsafe) x2,f =  0.1454208E-004-0.2180923E-002
581:WARNING from PE   509: (Task 509) MARBL ERROR (marbl_co2calc_mod:drtsafe): bounding bracket for pH solution not found
581:WARNING from PE   509: (Task 509) MARBL ERROR (marbl_co2calc_mod:drtsafe): (marbl_co2calc_mod:drtsafe) dic =  0.2000841E+004
581:WARNING from PE   509: (Task 509) MARBL ERROR (marbl_co2calc_mod:drtsafe): (marbl_co2calc_mod:drtsafe) ta =  0.2403694E+004
581:WARNING from PE   509: (Task 509) MARBL ERROR (marbl_co2calc_mod:drtsafe): (marbl_co2calc_mod:drtsafe) pt =  0.4128276E+000
581:WARNING from PE   509: (Task 509) MARBL ERROR (marbl_co2calc_mod:drtsafe): (marbl_co2calc_mod:drtsafe) sit =  0.6506462E+001
581:WARNING from PE   509: (Task 509) MARBL ERROR (marbl_co2calc_mod:drtsafe): (marbl_co2calc_mod:drtsafe) temp =  0.2542597E+002
581:WARNING from PE   509: (Task 509) MARBL ERROR (marbl_co2calc_mod:drtsafe): (marbl_co2calc_mod:drtsafe) salt =  0.3584534E+002
581:WARNING from PE   509: (Task 509) MARBL ERROR (marbl_co2calc_mod:comp_htotal): Error reported from drtsafe
```

I was thinking about two separate tests:

1. try `max_bracket_grow_it = 4` (I know 5 works and 3 doesn't)
2. switch to `nstep` output and see what things look like right before the failure; at this point, all I can tell is that the error occurs in the first day though I wouldn't be surprised if it was in the first time step.

Before I get to that, I realized that I forgot to update the dummy MARBL interface to include `bot_flux_to_tend` on the interface class... I'd like to get the CI passing, though I'm having trouble getting a compute node on cheyenne to test the build right now

Issue #204, opened 2017-10-30
add nml based option to ciso to turn off fractionation and 14c decay


klindsay28:
purpose is to be able to compare ciso tracers to base tracers
with fractionation and 14c decay off, and appropriate initializtion,
the ciso tracers should be identical to the (non-isotopic) base tracers
i.e., di13c should be the same as dic, ...

PR #386, opened 2021-08-30
Replace dzr_loc with bot_flux_to_tend

mnlevy1981:
Reworked some of the logic around computing `sed_denitrif` and `other_remin` to
remove the `dzr_loc` (this should have been done as part of #381)


mnlevy1981:
@klindsay28 -- the last few MOM runs we've looked at have used this branch, and it looks like it keeps O2 from going off the rails. Are there other diagnostics we should look at before merging? Any comments on the code itself?

klindsay28:
It looks like these changes should also be applied to `compute_particulate_terms` in `marbl_ciso_interior_tendency_mod.F90`.

I suspect that if #204 were implemented, this would have been caught.

klindsay28 (approved):
Looks good to me. Thanks!

PR #388, opened 2021-09-07
Migrate from Travis to gh-actions

mnlevy1981:
Travis CI doesn't seem to be configured correctly, so the past few PRs were tested by hand. Following the guidance of [ESMCI/cime](https://github.com/ESMCI/cime/blob/master/.github/workflows/srt.yml), this PR is attempting to use github actions for continuous integration instead of Travis.

Note that Travis let me rely on a very outdated version of Ubuntu, so I was testing with gfortran 4.8 and an old openmpi version -- on my laptop, the `--mpitasks 2` fails at build with

```
  148 |     call MPI_Bcast(logical_to_bcast, 1, MPI_LOGICAL, root, MPI_COMM_WORLD, ierr)
      |                   1
......
  166 |     call MPI_Bcast(int_to_bcast, 1, MPI_INTEGER, root, MPI_COMM_WORLD, ierr)
      |                   2
Error: Type mismatch between actual argument at (1) and actual argument at (2) (LOGICAL(4)/INTEGER(4)).
```

```
  129 |     call MPI_Bcast(str_to_bcast, len(str_to_bcast), MPI_CHARACTER, root, MPI_COMM_WORLD, ierr)
      |                   1
......
  166 |     call MPI_Bcast(int_to_bcast, 1, MPI_INTEGER, root, MPI_COMM_WORLD, ierr)
      |                   2
Error: Type mismatch between actual argument at (1) and actual argument at (2) (CHARACTER(*)/INTEGER(4)).
```

I found a fix for this (replacing `include 'mpif.h'` with `use mpi`), which will be coming in a later commit to this PR but this may mean we can only build the model serially using the much older `gfortran` (which I believe is still needed for the matlab / python drivers, although I haven't tested that with a newer compiler).


mnlevy1981:
closed in favor of #389

PR #403, opened 2022-05-10
Improve diagnostics

### LINKED ISSUE #400: Output FEISTY forcing directly
### Summary of issue

It would be great if POP and MARBL could combine to generate FEISTY forcing directly, rather than FEISTY users post-processing existing diagnostics to generate forcing file.

****

### Details

To run FEISTY, we use the following forcing fields:

1. Average temperature over top 150m of column
2. Temperature at bottom of column
3. POC flux to seafloor
4. Total mesozooplankton biomass over top 150m of column
5. Total mortality loss of mesozooplankton over top 150m of column

We need to add these to POP:

1. Average temp over top 150m
2. Temperature at bottom of column
3. Integral of `((zoo))C` over 150 m for all `((zoo))` tracers (we just want this for mesozoo)

We need to add these to MARBL:

4. POC flux at KMT
5. Integral of `zoo_loss` over top 150 m

#### Notes

1. POP can already output tracer integrals over 100m (e.g. `zooC_zint_100m`), and MARBL can compute `zoo_loss_zint_100m` so for first test we can use these built-in variables (and compute average temperature over top 100m). This would still require two new POP diagnostics: `TEMP_mean_100m` and `TEMP_bottom`. We also need `POC_flux_bottom` in MARBL.
2. At the end of Section 2.1 in [her paper](https://www.sciencedirect.com/science/article/abs/pii/S0079661118301952), Colleen actually uses 100m means / integrals. Do we really want 150m, or is 100m sufficient?

### LINKED ISSUE #401: photoC_NO3_TOT_zint_100m diagnostic not computed
In recent MOM vs POP MARBL comparison runs, I noticed that `photoC_NO3_TOT_zint_100m` is all zeros.

Digging into the code and commit history, I see that I added a declaration for `photoC_NO3_TOT_zint_100m` in e7deac50, but did not add code to compute the diagnostic.

I guess we haven't looked at this diagnostic. It is not a CMIP diagnostic, while the related diagnostic `photoC_NO3_TOT_zint` is.

I'm not sure if we should remove `photoC_NO3_TOT_zint_100m` or add code to compute it.

I'm leaning towards the former, given that this issue hasn't been noticed in the 4 years since the diagnostic was added.

mnlevy1981:
This PR addresses two small issues with diagnostics:

1. fixes #401 (`photoC_NO3_TOT_zint_100m` was being defined but never computed)
2. resolves #400 (`pocToFloor` is now output in the `medium` stream by default)

For the first, I opted to compute `photoC_NO3_TOT_zint_100m` rather than remove it from the diagnostic list.

There are two small issues remaining to address:

* I still need to update the baseline file to avoid an error when comparing results from `tests/regression_tests/call_compute_subroutines/`
* There is a CI test failure in building the documentation that I think stems from an incompatibility between the version of `sphinx` we are using for the docs and a major version update of `jinja2`.

I'll remove draft status from this PR after fixing those two minor issues.


mnlevy1981:
After talking with @klindsay28, we thought the best plan would be to

1. Bring in the CI changes in a separate PR (#404)
2. Discuss the two other changes in this PR [on Tuesday](https://github.com/marbl-ecosys/MARBL/wiki/2022-MARBL-Dev-team-meetings#may-17-2022)

I'm going to close this PR, and will open a new one (or two) next week, depending on how the conversation goes.

PR #419, opened 2023-02-22
Updates from mom6 driver

### LINKED ISSUE #417: Generalize marbl_surface_flux_output_type
MOM6 wants the full 3D chlorophyll field instead of just surface chlorophyll when computing absorption. In talking with @klindsay28, we think the best plan forward would be to

1. rename `marbl_surface_flux_output_type` -> `marbl_output_for_GCM_type` (also rename `sfo_cnt` -> `output_count` and add `num_levels`; may also need to split `num_elements` -> `num_elements_surface_flux` and introduce `num_elements_interior_tendency`)
2. rename `marbl_single_sfo_type` -> `marbl_single_output_type` (also rename `forcing_field` -> `forcing_field_0d` and introduce `forcing_field_1d`; though that adds more work to the resolution of #190 )
3. Introduce `interior_tendency` equivalents of the current `surface_flux_output` variables (and indices); have `interior_tendency_compute()` add chlorophyll in the correct spot
4. In `marbl_surface_flux_output_indexing_type`, rename `totalChl_id` to `surf_totalChl_id` (and we'll use `totalChl_id` in the interior tendency equivalent)

I think renaming `forcing_field` -> `forcing_field_0d` and `totalChl_id` -> `surf_totalChl_id` will require updates on the GCM side of things. I don't plan on updating POP to use this tag, but am happy to do so if it would be beneficial (POP would continue to use the surface chlorophyll, and I doubt we will be supporting POP long enough for a future PR to introduce a feature that is needed in that model). @maltrud and @mark-petersen this could have implications in E3SM / MPAS-O as well, so I want to make sure you're aware of this issue ticket.

mnlevy1981:
The big feature is generalizing the `surface_flux_output_type` so MARBL can provide the GCM with fields from `interior_tendency_compute()` as well as `surface_flux_compute()`; this new functionality is used to allow the option of returning the 3D chlorophyll field (POP only needed the surface chlorophyll, but MOM needs the 3D field).

I also added an option to check consistency in the PAR forcing field, but it is turned off by default because there is a huge performance hit.


mnlevy1981:
@klindsay28 and I reviewed this PR on March 8th; 0c46cd5, 1828803, and 75852f1 address all the small changes. The only outstanding item from the code review is that `check_forcing()` is triggering a ton of warnings in a C compset,  and the additional output to stdout is a significant hindrance to performance. Before this is merged, we need to figure out what is going wrong with the C compset forcing; once those warnings are reduced I'll have one more commit to this PR where I remove the `lcheck_forcing` logical flag (we'll check the forcing every time)

mnlevy1981:
Per further discussion with @klindsay28, we're going to merge this as is and address the problems in `check_forcing()` separately.

Issue #320, opened 2018-09-11
grazing type should be a subset of zooplankton type


mnlevy1981:
Rather than have `grazing(prey_cnt, pred_cnt)`, the `zooplankton_type` should have a `grazer_prey_cnt` variable and then an allocatable `grazing(:)` structure. Basically, this replaces `grazing(prey_cnt, pred_cnt)` with `zooplankton(zoo_cnt)%grazing(:)` (Per #317 these names will likely change).

PR #423, opened 2023-04-19
initial commit for 4p2z settings

kristenkrumhardt:
This is my first attempt at creating the yaml file for the 4p2z settings.


mnlevy1981:
**Progress so far**

1. The `marbl.settings` file generated by `MARBL_tools/MARBL_generate_settings_file.py` pointing at the new `4p2z` JSON file (and using `CESM_x1` for the grid) produces bit-for-bit identical results in the `call_compute_subroutines` test when compared to the same test reading the `marbl_in` file @kristenkrumhardt has been using for her 4p2z runs, e.g. `/glade/work/kristenk/cesm_work/cesm2.2_scope_cases/g.e22.GOMIPECOIAF_JRA-1p4-2018.TL319_g17.4p2z.001branch/CaseDocs/marbl_in` (note that I needed to generate an updated IC file to include `cocco` tracers and split `zooC` into `microzooC` and `mesozooC`; I also updated both settings files to turn off tracer restoring)
2. I've merged the `update_zoo_diags` branch into this branch; there were some diagnostics that were named `*_zoo*` that were effectively `*_zootot*`, but now that we can run with multiple zooplankton we should really be providing them as per-zoo fields.

The next step will be to merge `development` into this branch, and run the same tests. There are also two points we should discuss before accepting this PR...

**Discussion Topics**

1. `update_zoo_diags` also includes a change to `picpoc` that is not BFB

```diff
diff --git a/src/marbl_interior_tendency_mod.F90 b/src/marbl_interior_tendency_mod.F90
index e6017ef..398b42c 100644
--- a/src/marbl_interior_tendency_mod.F90
+++ b/src/marbl_interior_tendency_mod.F90
@@ -1452,7 +1452,8 @@ subroutine compute_autotroph_calcification(km, autotroph_local, temperature, car
           picpoc = max(0.0_r8, -0.0136_r8 * CO2(:) + picpoc(:) + 0.21_r8)

           !P-limitation effect (CaCO2/organicC ratio increases when P limitation term is low)
-          picpoc = max(0.0_r8, -0.48_r8 * Plim(auto_ind,:) + picpoc(:) + 0.48_r8)
+          ! max out picpoc at one to prevent crashing due to "ballast exceeds POC" error
+          picpoc = min(1., max(0.0_r8, -0.48_r8 * Plim(auto_ind,:) + picpoc(:) + 0.48_r8))

           !multiply cocco growth rate by picpoc to get CaCO3 formation
           CaCO3_form(auto_ind,:) = picpoc(:) * photoC(auto_ind,:)
```

I'd like to know if we would expect to see answer changes in the current 3p1z setup, or if `picpoc` is only used in explicit calcification so it will only effect runs with `cocco`

2. The current way we handle grazing requires the `pred_ind` in `grazing_relationship_settings(prey_ind, pred_ind)` to match the index of `zooplankton_settings(:)` referring to the predator. This can be seen in [`marbl_interior_tendency_mod.F90`](https://github.com/marbl-ecosys/MARBL/blob/75852f1/src/marbl_interior_tendency_mod.F90#L1899-L1917), where we need the biomass of the predator and use `zooplankton_local%C(pred_ind,k)`. Addressing #320 would remove this problem, but in the meantime we need `grazing_relationship_settings(4,1) = "null"` and `grazing_relationship_settings(4,2) = "Grazing of microzooplankton by mesozooplankton"` because `grazing_relationship_settings(4,1) = "Grazing of microzooplankton by mesozooplankton"` implicitly uses `microzoo` biomass instead of `mesozoo` biomass.

I am happy waiting until we fix #320 and letting this take care of itself, but an alternative approach would be to add `pred_ind` to the `grazing_relationship_settings_type`; we could then collapse `grazing_relationship_settings` to a 1D array (there are a few changes in this PR that are meant to handle the fact that there are 7 grazing relationships in the 4p2z setup but we need to allocate a 4x2 array to store them all)

**Next Steps**

Next up, I'll merge development into this branch; it looks like there are conflicts in a couple of files to resolve, and I'll also need to add the following new variables to the 4p2z settings file:

```diff
diff --git a/defaults/settings_latest.yaml b/defaults/settings_latest.yaml
index 87e50c0..2f105ce 100644
--- a/defaults/settings_latest.yaml
+++ b/defaults/settings_latest.yaml
@@ -232,6 +232,12 @@ general_parms :
       units : unitless
       datatype : logical
       default_value : .true.
+   lcheck_forcing :
+      longname : Control whether consistency checks are performed on forcing input (useful for debugging)
+      subcategory : 2. config flags
+      units : unitless
+      datatype : logical
+      default_value : .false.
    lecovars_full_depth_tavg :
       longname : If .false., MARBL will recommend truncating the column for some diagnostics
       subcategory : 2. config flags
@@ -283,6 +289,12 @@ general_parms :
       units : unitless
       datatype : logical
       default_value : .false.
+   bftt_dz_sum_thres:
+      longname : MARBL will abort if abs(1 - sum(bot_flux_to_tend)) exceeds this threshold
+      subcategory : 4. general parameters
+      units : unitless
+      datatype : real
+      default_value : 1.0e-14
    Jint_Ctot_thres_molpm2pyr :
       longname : MARBL will abort if abs(Jint_Ctot) exceeds this threshold
       subcategory : 4. general parameters
@@ -1008,6 +1020,14 @@ PFT_derived_types :
             default_value :
                default : 1e34
                ((zooplankton_sname)) == "zoo" : 0.4
+         basal_respiration_rate_per_day :
+            longname : Basal respiration rate
+            subcategory : 11. zooplankton
+            units : 1/day
+            datatype : real
+            default_value :
+               default : 1e34
+               ((zooplankton_sname)) == "zoo" : 0.0
          loss_thres :
             longname : Zoo concentration where losses go to zero
             subcategory : 11. zooplankton
```

PR #427, opened 2023-05-18
Update defaults for tracer_restore_vars(:)

mnlevy1981:
Instead of restoring PO4, NO3, SiO3, ALK, and ALK_ALT_CO2 when GRID is either CESM_x3 or CESM_x1, those five are set when a new variable (USE_CESM_RESTORE) is set to TRUE. TRUE is the default value, but I have a MOM_interface update that will use FALSE instead, and then I don't need the kludge in MOM's MARBL_tracers module to turn off restoring (which is good, because we want to start using restoring in the single column runs that run through solo_driver)

### DIFF
--- a/MARBL_tools/MARBL_generate_settings_file.py
+++ b/MARBL_tools/MARBL_generate_settings_file.py
@@ -24,11 +24,13 @@ optional arguments:
   -f DEFAULT_SETTINGS_FILE, --default_settings_file DEFAULT_SETTINGS_FILE
                         Location of JSON-formatted MARBL configuration file
                         (default: $MARBLROOT/defaults/json/settings_latest.json)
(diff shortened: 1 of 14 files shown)


mnlevy1981:
I chose to set this up with the default for `restore_for_CESM` set to `True` because setting it to `False` would require a change to POP2's `buildnml` script the next time we wanted to update MARBL... but it probably makes more sense to have the default be `False` despite that additional work.

I also think `restore_for_CESM` is a terrible name for this variable but I wasn't sure what property ties `PO4`, `NO3`, `SiO3`, `ALK`, and `ALK_ALT_CO2` together that would maybe provide a clearer name.

mnlevy1981:
Closing this in favor introducing a new `GRID` option for MOM6 and duplicating `GRID == "CESM_x1"` defaults for that grid when necessary

PR #441, opened 2023-08-31
Add abiotic tracers

### LINKED ISSUE #444: MARBL_settings_class assumes units in input_file match YAML, not unit_system
If I am running CESM+MOM6, I can use `user_nl_marbl` to change variable values in `marbl_in`. Since MOM6 is in `mks`, I want to use `mks` values in `user_nl_marbl`, but the python script that generates `marbl_in` will assume `user_nl_marbl` will provide values in the units defined in the settings YAML file (which is cgs for the most part).

Here's a concrete example -- `parm_POC_diss` is defined in the YAML as

```yaml
   parm_POC_diss :
      defined_if : base_bio_on
      longname : Base POC dissolution length scale
      subcategory : 5. general parameters (dissolution)
      units : cm
      datatype : real
      default_value : 100e2
```

In MOM6, this default value is correctly appears in `marbl_in` as `parm_POC_diss =   1.0000000000000000e+02` because we convert the default value to mks. However, if I set

```
parm_POC_diss = 150
```

in `user_nl_marbl`, then `marbl_in` contains `parm_POC_diss =   1.5000000000000000e+00` rather than the desired `parm_POC_diss =   1.5000000000000000e+02`. The fix is to return a scale factor of 1. to `_translate_JSON_value()` if the variable came from an input file.

mnlevy1981:
Port the `abio_dic_dic14_mod.F90` module from POP to MARBL. We should be able to run with the two abiotic tracers without turning on what the base (non-isotopic) MARBL tracers.


mnlevy1981:
Currently we are using same atmospheric CO2 forcing from GCM for biotic and abiotic tracers; while the majority of runs will likely use the same value for both, we do want to support running both modules with different CO2 forcings (Keith L gave the example of using emissions-driven CO2 for base biotic tracers in a coupled run but something different for abiotic)

mnlevy1981:
If `abio_on = .true.` and `base_biotic_on = .false.`, then we need `autotroph_cnt` and `zooplankton_cnt` set to 0 (may as well set `max_grazer_prey_cnt` as well, for completeness) so that `setup_local_tracers()` doesn't try to set tracers that don't exist. This issue is cropping up now because I just moved the call to `setup_local_tracers()` out of the `if (base_biotic_on)` block.

mnlevy1981:
commits [6f552e9](https://github.com/marbl-ecosys/MARBL/pull/441/commits/6f552e92e479ac3ff665853304be4bae51221fe0) and [a8bb2a7](https://github.com/marbl-ecosys/MARBL/pull/441/commits/a8bb2a7b56a3657a0b8fc7e4a68251212f5853d1) reverted some changes in settings YAML files that could have been caught by a regression test for `MARBL_generate_settings.py` that runs for each settings JSON file and makes sure the resulting settings file matches a baseline (in this case, there are expected changes but the YAML file also introduced several unexpected changes). I think that's a simple addition to `run_test_suite.sh` but if it turns into something more complicated I'll open a separate issue / PR for it

mnlevy1981:
The tests that fail in 5a1f06e pass on my laptop, but fail on cheyenne (with gnu) -- I'll investigate further on that machine, it's not clear what changed in that commit to cause problems.

PR #457, opened 2024-03-01
generate_settings_file sets tracer module logicals

mnlevy1981:
Command line arguments for `MARBL_generate_settings_file.py` determine value of `base_bio_on` and `abio_dic_on`: by default, `base_bio_on=T` but running `--exclude-base-bio` sets it to F. Similarly, `abio_dic_on=F` but running `--include-abio-dic` sets it to T.


mnlevy1981:
This PR will let CESM users specify whether to use the base biotic tracers or abiotic DIC tracers (or both) via a variable in `env_run.xml` (that can be set differently for different compsets) and generate the proper `marbl_in`

PR #462, opened 2024-05-09
Add diag_mode to default diagnostics

mnlevy1981:
Introduce new `diag_mode` option to `MARBL_generate_diagnostics_file.py`; when `diag_mode = 'none'`, output diagnostic file should be `'never_{op}'` for all variables. When `diag_mode = 'minimal'`, only variables with `diag_mode = 'minimal'` in `diagnostics_latest.json` have non-never output frequencies. When `diag_mode is 'full'`, variables with `diag_mode = 'minimal'` or `'full'` will have non-never output frequencies.

### DIFF
--- a/MARBL_tools/MARBL_diagnostics_file_class.py
+++ b/MARBL_tools/MARBL_diagnostics_file_class.py
@@ -53,10 +53,11 @@ class MARBL_diagnostics_class(object):
                 diags_to_delete.append(diag_name)
                 continue
 
-            #    iii. frequency and operator should always be lists
+            #    iii. frequency, operator, and diag_mode should always be lists
             if not isinstance(self.diagnostics_dict[diag_name]['frequency'], list):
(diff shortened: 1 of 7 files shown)


mnlevy1981:
Just to have this recorded somewhere, I tested this PR in CESM with ocean-only (MOM6) cases. Each was run for a single month.

| `MARBL_DIAG_MODE` | Size of `RUNDIR` |
| --- | --- |
| `none` | 18 GB |
| `minimal` | 19 GB |
| `full` | 29 GB |
| `test_suite` | 45 GB |

Notes
1. `test_suite` is only defined in `MOM_interface`; it writes monthly means of every diagnostic MARBL provides. For all the other values, `MARBL_DIAG_MODE` is used for `diag_mode` as defined in this PR
2. MOM6 writes all output in single precision, with one exception: when `diag_mode = test_suite`, the MARBL diagnostics (as well as the MARBL tracer-specific diagnostics computed in MOM6) are written in double precision

PR #465, opened 2024-07-18
Update marbl_ciso_surface_flux_mod.F90

jlinmod:
comment and eps_aq_g equation fix


mnlevy1981:
@jlinmod -- sorry for the delayed response. The Zhang et al. paper offers two different values for the temperature coeffficient when computing `eps_aq_g`. In the abstract, they use `-0.0049` as you did:

zhang abstract

However, in Figure 2 they use `0.0049` as in the current version of the code:

zhang fig 2

Further, the slope of the line in Figure 2 is positive, which makes us think that the figure caption is the correct value to use.

Rather than close this PR, could you revert line 155 to drop the minus sign? I think the clean-up you did to the comments is valuable and we'd be happy to bring those changes in.

PR #466, opened 2024-07-18
Update marbl_ciso_interior_tendency_mod.F90

jlinmod:
Fixed comments, eps_aq_g equation error (Zhang et al., 1995), and cell surface area unit conversion fix for calcifier (Popp et al., 1998).

### DIFF
--- a/src/marbl_ciso_interior_tendency_mod.F90
+++ b/src/marbl_ciso_interior_tendency_mod.F90
@@ -325,6 +325,8 @@ contains
        !-----------------------------------------------------------------------
        !   discrimination factors of carbone chemistry based on
        !   Zhang et al, 1995, Geochim. et Cosmochim. Acta, 59 (1), 107-114
+       !   Note: there is a discrepancy between the figure caption and the abstract of the paper, 
+       !   the typo is in the abstract, no negative 0.0049
        !
        !   eps = permil fractionation and alpha is the fractionation factor
        !   with eps =(alpha - 1) *1000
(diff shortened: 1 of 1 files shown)


mnlevy1981:
As in https://github.com/marbl-ecosys/MARBL/pull/465#issuecomment-2359478490, [line 355](https://github.com/marbl-ecosys/MARBL/blob/93cc51fe733abf33e19bc8d5c2c60e3eecbed1a6/src/marbl_ciso_interior_tendency_mod.F90#L355) was correct before you added the negative sign but if you revert that change then we'd be happy to update all of your comments.

Also note that your change to `cell_surf(auto_ind)` in [line 836](https://github.com/marbl-ecosys/MARBL/blob/93cc51fe733abf33e19bc8d5c2c60e3eecbed1a6/src/marbl_ciso_interior_tendency_mod.F90#L836) is inside an entire block of [commented-out code](https://github.com/marbl-ecosys/MARBL/blob/93cc51fe733abf33e19bc8d5c2c60e3eecbed1a6/src/marbl_ciso_interior_tendency_mod.F90#L828-L840); as the comment mentions, we don't know these values for explicit calcifiers yet, so those autotrophs continue to use the [defaults for small phytoplankton](https://github.com/marbl-ecosys/MARBL/blob/93cc51fe733abf33e19bc8d5c2c60e3eecbed1a6/src/marbl_ciso_interior_tendency_mod.F90#L848-L859) (where `cell_surf(auto_ind) = 87.6e-12_r8`). I would suggest reverting line 836 as well, and we'll make sure `cell_surf(auto_ind)` is set correctly when we are ready to use different values for all the `cell_` variables for explicit calcifiers.

PR #477, opened 2024-11-18
Add tracebacks in marbl_abio_dic_surface_flux_mod

mnlevy1981:
Errors raised in the carbonate solver were not being caught in the right place, leading to confusing error messages.


klindsay28:
LGTM

PR #464, opened 2024-05-23
Fe updates from UCI

mnlevy1981:
This branch contains the code changes used by the UCI group; it includes variable C:N as well as new iron forcing

PR #481, opened 2025-08-07
Force older docutils in environment

mnlevy1981:
CI tests were failing with

Could not import extension sphinx.builders.latex (exception: No module named 'docutils.utils.roman')

According to https://docutils.sourceforge.io/RELEASE-NOTES.html the docutils/utils/roman.py file was removed in 0.22, so we require something older


mnlevy1981:
I used `git cherry-pick` to bring in e589f7f60d0 from #464, hopefully that will make it easier to use `git merge` to bring the latest `development` back to that branch after this gets merged in

PR #483, opened 2025-09-07
Update marbl_settings_mod.F90 80 yr for DOMr_reminR_photo

jlinmod:


mnlevy1981:
Thanks for this contribution! I'd like to get #464 merged before we investigate the effect of this in CESM3, but it looks promising.

PR #485, opened 2025-10-01
Updates to fix nvhpc compiler issues

mnlevy1981:
To run the standalone tests on derecho, we want to use nvhpc/25.9 (this includes a bug fix we reported where nvhpc/25.7 was not handling associate statements correctly).

The compiler also rightly complained about an instance of a missing kind number when using a real variable (although 1.0 is exactly representable so there shouldn't be any difference between 1.0 and 1.0_r8).

Thirdly, I replaced a couple of where statements with the do-loop if-statement combination to avoid dividing by zero. It is unfortunate that the compiler doesn't handle this very well, but it is not a very invasive change.

### DIFF
--- a/src/marbl_interior_tendency_mod.F90
+++ b/src/marbl_interior_tendency_mod.F90
@@ -1541,7 +1541,7 @@ contains
 
           !P-limitation effect (CaCO2/organicC ratio increases when P limitation term is low)
           ! max out picpoc at one to prevent crashing due to "ballast exceeds POC" error
-          picpoc = min(1., max(0.0_r8, -0.48_r8 * Plim(auto_ind,:) + picpoc(:) + 0.48_r8))
+          picpoc = min(1.0_r8, max(0.0_r8, -0.48_r8 * Plim(auto_ind,:) + picpoc(:) + 0.48_r8))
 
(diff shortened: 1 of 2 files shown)


mnlevy1981:
Note: I am trying to create a small example problem highlighting the issue with `where` statements in the nvhpc compiler, but my simple example doesn't create the same error I was seeing in the MARBL test suite. I expected

```fortran
program where_statements

  real, dimension(3,3) :: old_vals, new_vals
  integer :: i

  old_vals(:,:) = 2.0
  old_vals(2,:) = 0.0

  do i=1,3
    where (old_vals(i,:) > 0.0)
      new_vals(i,:) = 1.0 / old_vals(i,:)
    else where
      new_vals(i,:) = 0.0
    end where
    print*, new_vals(i,:)
  end do


end program where_statements
```

to fail, but even with the same `-O2 -Mfree` compiler flags we use to build MARBL it runs fine. I don't think it's worth spending the time to further investigate / understand why I needed to modify `compute_autotroph_nutrient_uptake()` in this PR, although I can think of a few debug statements to include that might help highlight the issue.

klindsay28 (approved):
LGTM

PR #486, opened 2025-11-06
Add new burial diagnostics

mnlevy1981:
Three derivatives used to compute burial coefficients when `ladjust_bury_coeff = .true.` are now available as diagnostics regardless of the value of `ladjust_bury_coeff`. They are not included in the default diagnostic set but will be useful for computing the burial coefficients offline for GCMs that do not provide the global operators necessary to adjust the burial coefficients on the fly.

### DIFF
--- a/defaults/diagnostics_latest.yaml
+++ b/defaults/diagnostics_latest.yaml
@@ -849,6 +849,33 @@ pfeToSed :
    frequency : medium
    operator : average
    diag_mode : minimal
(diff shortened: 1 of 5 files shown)


mnlevy1981:
I expect some regression test failures because these new diagnostics are not included in the baselines for the `call_compute_subroutines` tests; I will create new baselines after the PR is reviewed and we are happy with the changes, and will not merge this until the CI tests all pass.

